Issue #24, opened 2021-02-12
Tests pass with unhandled rejections in node 14 and node 12


haikyuu:
Tests do pass but unhandled rejections are there in console. You can see that [here](https://travis-ci.org/github/haikyuu/linkedom/jobs/758659588)

WebReflection:
then tests don't pass, right? ... interesting, no idea why I wouldn't see this in node 15, will check it, thanks.

WebReflection:
found it ... neither node 14 nor 12 have `Event` interface, and the `@ungap` polyfill relies on a global document ... bummer

haikyuu:
found that too but I was unsure what to do with ungap. Maybe just copy it in the source and refactor it to use a document parameter?

WebReflection:
naa, I've dropped one dependency, which feels *way* better 🥳

haikyuu:
😆  I like that commit message `Poorlyfill`

PR #25, opened 2021-02-12
Tests pass with unhandled rejections

haikyuu:
This PR showcases #24

I only added 14 to showcase the issue but I think it's better to add 12 as well. 15 is no LTS so it's not representative of real usage.
- [x]  add node 14 version to travis CI
- [ ] Fix unhandled rejections

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: node_js
 node_js:
   - stable
+  - v14
+  - v12
 git:
   depth: 1
 branches:


WebReflection:
Thanks!

  * please rebase to latest
  * please squash commits as one
  * please do not include semver details, just `v14` and `v12` so that we know LTS versions are fine, no matter which one these are

WebReflection:
uhm ... travis ain't happy with `sh` command ... trying to fix that

WebReflection:
do you mind rebasing again so we are sure both 14 and 12 run? stable is green now, thanks

WebReflection:
actually, I'll check that, will merge, thanks 👋

haikyuu:
Thank you 👍🏽

WebReflection:
no, thank you! I wouldn't have known `@ungap/event` was actually completely useless in there ... also, please update to `0.4.20` for best perf and less errors (there was an SVGElement regression) 👋

PR #27, opened 2021-02-15
[TASK] Add options property on HTMLSelectElement

SamBrishes:
Hellow,

I'm currently implementing the linkedom package in my test environment and noticed, that the `HTMLSelectElement.options` property is missing (among a few other ones). I implemented an inaccurate version of it for my own purposes, "inaccurate" since it returns a `NodeList` instance instead of the correct `HTMLOptionsCollection` (living standard) / `HTMLCollection`. Furthermore, I included a test as well and at least on Ubuntu@WSL it writes a bunch of green 100's.

Thanks for your work, really appreciate it.

Sincerely,
Sam.

### DIFF
--- a/cjs/html/select-element.js
+++ b/cjs/html/select-element.js
@@ -3,6 +3,7 @@ const {registerHTMLClass} = require('../shared/register-html-class.js');
 const {booleanAttribute} = require('../shared/attributes.js');
 
 const {HTMLElement} = require('./element.js');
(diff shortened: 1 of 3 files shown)


WebReflection:
@SamBrishes amazing! Thank you for this clean MR, I guess one day I'll put `selectedIndex` in too, but before that I need `<option>` to be selectable, although I love these kind of incremental updates :heart:

PR #56, opened 2021-04-28
Fix set outerHTML()

minas90:
The new implementation should properly handle things like:

```
el.outerHTML = 'hello';
el.outerHTML = '<p>hello</p> world'
```

The current implementation will throw an exception for the first case and will set it to `'<p>hello</p>' ` instead of `'<p>hello</p> world'` in the 2nd case.`

### DIFF
--- a/cjs/interface/element.js
+++ b/cjs/interface/element.js
@@ -197,7 +197,7 @@ class Element extends ParentNode {
   set outerHTML(html) {
     const template = this.ownerDocument.createElement('');
     template.innerHTML = html;
-    this.parentNode.replaceChild(template.firstElementChild, this);
+    this.replaceWith(...template.childNodes);
   }
   // </contentRelated>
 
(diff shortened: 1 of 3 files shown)


WebReflection:
Thanks, but two things:
  1. the source of truth is the ESM folder, not the CJS one
  2. I'd like to have a test that covers all cases `<p>other</p>`, `other`, `<p>other</p> more`

WebReflection:
almost there ... to test that code coverage is still 100% and everything work, `npm i` and `npm run build` would do.

tests are in the `test` folder, you can pollute the `test/interface/named-node-map.js` if you like, I wouldn't mind, as long as I am sure there won't be regressions in the future.

thank you!

minas90:
Hi Andrea,
I added a test file, sorry for a bit messy commits. I have a different project and test structure, so I got confused haha.

minas90:
Thanks for the amazing project. It's up to 10x faster compared to jsdom for my use case. I will contribute more small changes later to make it come closer to the spec.

WebReflection:
@minas90 thank you! it's up and running 👋

PR #57, opened 2021-04-29
unescape text nodes for innerHTML and outerHTML

minas90:
`get innerHTML` and `get outerHTML` should unescape text nodes, so `<div>"hello"</div>` should not become `<div>&quot;hello&quot;</div>`

### DIFF
--- a/cjs/interface/element.js
+++ b/cjs/interface/element.js
@@ -462,7 +462,7 @@ class Element extends ParentNode {
           break;
         case TEXT_NODE:
         case COMMENT_NODE:
-          out.push((isOpened ? '>' : '') + next);
+          out.push((isOpened ? '>' : '') + unescape(next));
           isOpened = false;
           break;
       }
--- a/esm/interface/element.js
+++ b/esm/interface/element.js
@@ -463,7 +463,7 @@ export class Element extends ParentNode {
           break;
         case TEXT_NODE:
         case COMMENT_NODE:
-          out.push((isOpened ? '>' : '') + next);
+          out.push((isOpened ? '>' : '') + unescape(next));
           isOpened = false;
           break;
       }
--- a/test/interface/element.js
+++ b/test/interface/element.js
@@ -9,7 +9,8 @@ let div = document.querySelector('div');
 div.firstChild.outerHTML = 'hello';
 assert(div.firstChild.toString(), 'hello');
 
-div.innerHTML = '<span></span>'
+div.innerHTML = '<span>"hi</span>';
+assert(div.firstChild.toString(), '<span>"hi</span>');
 div.firstChild.outerHTML = '<p>hello</p>';
 assert(div.firstChild.toString(), '<p>hello</p>');
 


WebReflection:
Tests failing ... but I have a dejavu here, as this was already fixed and tested/covered?

minas90:
I guess it was fixed for a special case, when it's just a text node. 
Here is the other fix at the end of this line.
`html.push(node.nodeType === ELEMENT_NODE ? node.toString() : unescape(node));`

Yeah, I guess tests need some rework, I will work on that later today.

WebReflection:
The toString doesn't escape though ... so it's not super clear why you see that 🤔

minas90:
Btw there is a similar issue with attributes, but it's the opposite, eg `<a href="https://w.com?k1=v1&amp;id=14` should NOT be unescaped to `<a href="https://w.com?k1=v1&id=14` as it may break urls in some cases.
Also `` should NOT become ``.
And also order of attributes is not persisted.
I will try to fix these issues as they break html and make it invalid.

minas90:
> The toString doesn't escape though ... so it's not super clear why you see that 🤔

The code is inside ` get innerHTML`, so eg if it's `<p>"hello</p>` and you do  `.innerHTML` you will get properly unescaped `"hello`, but if it's `<p><span>"hello<span></p>` and you do innerHTML for `p`, you will get `<span>&quot;hello<span>` and it's the same for `outerHTML`

WebReflection:
> Also  should NOT become .

That requires an exception rule for attributes ... and I question why you pass empty attributes instead, as it's not a too-interesting use case here, imho.

Bear in mind this project has zero interest in passing all HTML standard rules, there's JSDOM for that.

> And also order of attributes is not persisted.

And likely never will due technical choices, but this is another not interesting use case, and at least on parsing the order remains, on setting, adding, it's always at the top of the element for performance reasons.

minas90:
> That requires an exception rule for attributes ... and I question why you pass empty attributes instead, as it's not a too-interesting use case here, imho.

There are some required attributes like `alt` for `img`, which is empty in many websites. eg ``. I have dozens of tests on my project which are broken when I switch from jsdom because of that.

> Bear in mind this project has zero interest in passing all HTML standard rules, there's JSDOM for that.

Yes, I read about that :) but I think if it's like 60% compliant to the spec now, with very small changes it can be like 99% compliant to the spec ( of course if we don't consider lived nodes). So it can be drop in replacement for jsdom for majority of the cases.
 
> And likely never will due technical choices, but this is another not interesting use case, and at least on parsing the order remains, on setting, adding, it's always at the top of the element for performance reasons.

I will take a look, it's nor persisted when you do `get innerHTML` or `get outerHTML` and majority of my project tests are broken because of that.

WebReflection:
> only some attributes can have no value, for all others it should stay as it is, even if it's an empty string

like I've said, we need an exception rule, probably per element ... or to keep it initially simple, just a list of well known attributes that can be used as "*boolean*" so that nothing changes/happens if an attribute name is not part of this list.

> with very small changes it can be like 99% compliant

I want to keep performance a priority too, or this project won't make sense to exist anymore, I hope that's reasonable.

> I will take a look

It's the same when parsing documents, it should be the same when injecting content ... but the order of attributes won't ever be guaranteed, so you need to flip some test case (reverse attributes order via a regexp and you should be good).

I've done that myself for few of my libraries ... the attributes are set as first node, or the linked list makes no sense anymore.

minas90:
> like I've said, we need an exception rule, probably per element ... or to keep it initially simple, just a list of well known attributes that can be used as "_boolean_" so that nothing changes/happens if an attribute name is not part of this list.

Yes,  I think it's a good thing to do, otherwise html will be invalid.

> I want to keep performance a priority too, or this project won't make sense to exist anymore, I hope that's reasonable.

Of course, there are things which will not affect performance at all and some other things will affect, but it will negligible. Eg unescape is very quick operation and it also reduces size.

> It's the same when parsing documents, it should be the same when injecting content ... but the order of attributes won't ever be guaranteed, so you need to flip some test case (reverse attributes order via a regexp and you should be good).

I already modified some basic tests, but it will be a pain, because I frequently add real world html documents as tests every day. It doesn't make sense to modify thousands of documents with hundreds of thousands of attributes. 


I don't have an idea what happens when I add `unescape(next)`, some tests start failing because of uncovered lines, but without the change somehow all the lines are covered. Do you know what can be the reason?

henricook:
This issue is touching on two of my main pain points, Linkedom's fast parsing is ready to rescue my slow project if these are solved.

- `&nbsp;` and other entities seem to be rendered in strings (i.e. it comes out as <div> </div> instead of <div>&nbsp;</div>)
- Empty attributes are elided, I have genuine use cases for `<div data-default-value="">`

minas90:
Yeah empty attribute values is a very common use case, I ended up forking to fix the behaviour, otherwise I had to modify hundreds of lines of code in my project and 3 other dependencies I use.
I pushed the fix in another pull request, but tests are failing, because many functions and lines are not covered.

minas90:
&nbsp; and also attribute values are not encoded, which may cause serious issues which are hard to debug. For example href attribute which contains unicode chars or special chars may fail to load resources in some cases.

amb26:
I also support the idea that these escaping rules should be fixed for text nodes. Note that there is a similar issue with respect to comment nodes.
I don't mind much about the attribute issues.

amb26:
No worries, now I looked in the PR content I see that the comment issue will get fixed too via fallthrough.

amb26:
Why closed?

minas90:
@amb26 Because it will never be merged, the author clearly said that he doesn't want to follow the standards.

WebReflection:
> the author clearly said that he doesn't want to follow the standards

That's [not what I've said though](https://github.com/WebReflection/linkedom/issues/63#issuecomment-851528374), and the issue has been solved.

If you want a better output, current implementation gives you that. If you want boolean attributes with redundant `=""` all over, please use JSDOM instead, as easy as that. LinkeDOM goal is not to pass W3C tests, is to be a handy tool that does the right thing in the fastest possible way, considering real-world scenarios, ignoring when appropriate.

It's explained in the README, there's no need to be snarky about it.

WebReflection:
P.S. your MR broke tests in CI ... that's not how I roll, so it was never even considered, as broken tests, or lower code coverage than current 100%, doesn't make me even think to have a look at the code ... not sure what are your coding standards, but these are mine.

minas90:
> P.S. your MR broke tests in CI ... that's not how I roll, so it was never even considered, as broken tests, or lower code coverage than current 100%, doesn't make me even think to have a look at the code ... not sure what are your coding standards, but these are mine.

My MR didn't break anything, the code is completely fine and I also wrote all the necessary tests to cover the code.
Your tests don't cover 100% of the code, it's illusion. It's like a minefield, a single new line of code may reveal the core problem that majority of the code is not covered by the tests.

WebReflection:
@minas90 you deleted the branch promptly, there was a big ❌ on CI.

My code coverage is based on c8, so in case file a bug there.

I am not willing to discuss anything in these terms regardless, to this is a first warning, hoping I don't need to ban you in the future.

Have a nice day.

Issue #60, opened 2021-05-18
Contributor documentation: Running tests, working in IntelliJ


henricook:
Hi @WebReflection ,

The `how-to-contribute.md` suggested creating an issue for further documentation that I think might be useful.

I'm trying to get to grips with the project so I can maybe solve my own problems, let me tell you how I got started and maybe in the process of telling me what I should have done it'll become apparent what to add to the docs.

I want to look at the `toString`'ing of `Element` so I found a test file. I couldn't see any sections e.g. like in Jest (`it("test thing thing")`), I assumed this `assert` format is a stylistic thing but since I'm fixing one very defined problem for now I didn't want to spend time reading about it just yet. I just commented out the whole of `test/html/element.js` and added in one test for my issue at the top whilst I tried some things out:

```javascript
const {document} = parseHTML('<!DOCTYPE html><html id="html" class="live"><!--&lt;hello&gt;-->&lt;hello&gt;</html>');
document.documentElement.innerHTML = '<div>Foo&nbsp;&nbsp;&nbsp;Bar</div>'
assert(document.documentElement.toString(), '<div>Foo&nbsp;&nbsp;&nbsp;Bar</div>')
```

I usually work in IntelliJ, I tried to run the test from there but it can't make sense of the global.symbolFor stuff:

![image](https://user-images.githubusercontent.com/82004/118689983-8290c000-b7ff-11eb-804a-51a653d5e538.png)

So I went to the CLI and luckily `npm run test -- test/html/element.js` - a pretty standard syntax seemed to work. **It would be great to have test commands documented**.

I got a lot of output though, a lot of things that I probably don't need when I'm just iterating on one little test, a big table with these headers: `File                              | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s    `

So **it would be great to know how to not get all this stuff in the output whilst I'm focusing on a small task**. Of course I'd come back and run a full test suite afterwards. I appreciate it's probably important a lot of the time, but it's quite daunting to a newbie.

My test output is present about halfway through the big chunk of output from this command, so I can work with it for now. **It would also be helpful to know how you might add `console.log` type statements whilst debugging**, they don't seem to come out in the test output currently. I'd use the debugger, but IntelliJ has issues as above.

WebReflection:
is [this](https://github.com/WebReflection/linkedom/blob/main/test/assert.js#L1-L15) too complicated?

[...]

PR #67, opened 2021-06-04
Add how to run test instructions to README

henricook:
Complementing https://github.com/WebReflection/linkedom/issues/60#issuecomment-850813365

### DIFF
--- a/how-to-contribute.md
+++ b/how-to-contribute.md
@@ -22,14 +22,17 @@ Almost every `./esm` file, within its folder, is replicated in the `./test` fold
 
 Each test is executed twice: once for `linkedom` module, and a second pass for `linkedom/cached` export.
 
+Run the tests with: `npm run test`
+
 The very beginning of the test ensures everything is OK in *ESM* exports-land, it runs a quick benchmark to be sure everything works as expected and, eventually, keep performance under control, and finally to verify that all code has been covered.
 
-If a proposed change decrease overall code coverage, this won't be accepted until the score is back to 100%.
+If a proposed change decreases overall code coverage, this won't be accepted until the score is back to 100%.
 
 **However**, there are various parts that are already covered and won't likely benefit from extra tests. This is specially the case for *HTMLClasses accessors*, like the `HTMLImageElement#src` getter and setter, and all others I might have not fully augmented yet.
 
 In these contributions cases, you are not required to write extra tests, but please add `/* c8 ignore start */` and `/*c8 ignore stop */` around the proposed new code, or place new accessors within previously defined `c8` directives.
 
+To run only one test file (e.g. whilst developing) you can edit test/index.js and comment/uncomment folders appropriately.
 
 ## Testing - Helpers
 


how-to-contribute.md
```diff
@@ -22,14 +22,17 @@ Almost every `./esm` file, within its folder, is replicated in the `./test` fold
...
 The very beginning of the test ensures everything is OK in *ESM* exports-land, it runs a quick benchmark to be sure everything works as expected and, eventually, keep performance under control, and finally to verify that all code has been covered.
 
-If a proposed change decrease overall code coverage, this won't be accepted until the score is back to 100%.
+If a proposed change decreases overall code coverage, this won't be accepted until the score is back to 100%.
 
 **However**, there are various parts that are already covered and won't likely benefit from extra tests. This is specially the case for *HTMLClasses accessors*, like the `HTMLImageElement#src` getter and setter, and all others I might have not fully augmented yet.
 
 In these contributions cases, you are not required to write extra tests, but please add `/* c8 ignore start */` and `/*c8 ignore stop */` around the proposed new code, or place new accessors within previously defined `c8` directives.
 
+To run only one test file (e.g. whilst developing) you can edit test/index.js and comment/uncomment folders appropriately.
```
henricook, line 35: @WebReflection Is this the best way to run only one test file? For example in the scenario where I'm building some new functionality and want to run only the tests associated with that area of the code, on repeat, whilst I work. Probably even without code coverage to swell the output.

WebReflection: The hint about “no need to code cover” was meant to simplify contributions, yet *new* functionalities should have code coverage around, and likely not change previous tests, unless the regression is meant (see non decoding chars I am still planning to work on). Long story short, I’m not sure I’ve answered your question, but if CI is red it means I should consider a PR theat didn’t bother writing, or updating, tests, and I don’t look at red CI contributions, as I don’t have extra time to extra fix 100% code coverage for contributors, but hopefully contributors have such time, if interested in the project. I should definitely copy/paste this comment in the contributing file, but this is all I have so far.

henricook: Sure thanks. I think you're seeing a more complex question than I'm actually asking :-) Let me lay out the scenario I'm imagining:

1. I want to build a new feature, for a new type of HTML element. For arguments sake lets say it's the `<footer>` element
2. I make `html/footer-element.js`
3. I want to write some tests for my new file, `html/footer-element.js` so I create `test/html/footer-element.js`
4. Because the rest of the tests don't matter whilst I'm working on this one new file, I want to only run `test/html/footer-element.js` over and over again whilst I'm coding <-- how would I do this? is the question
5. I want the output to be brief and to the point, so i can easily see whether my new tests are passing or failing - which I think means no code coverage or other tests should be run each time <-- how would I do this? is the question
6. Once i've finished everything I would of course run the full test suite, code coverage analysis etc before creating a PR

WebReflection: Apologies I've read this waaaaaay too late ... so, I've landed [this MR](https://github.com/WebReflection/linkedom/pull/76) that basicaly solves your issue, or better, it allows you to test a single file, **as long as you require assert** in the process.

Would that fix all your questions?

henricook: Yep, perfect thanks!

WebReflection: Maybe putting c8 before node helps covering the file too

WebReflection: actually, `npm run cjs && npx c8 node test/html/anchor-element.js` could be pretty noisy, so never mind, `npm run cjs && node test/html/anchor-element.js` looks fine.

PR #68, opened 2021-06-06
Fix #62 - Follow specs for CharsetData un/escape

### LINKED ISSUE #62: HTML Entities cannot be outputted
For a large variety of use cases it is desirable to have HTML entities in Linkedom output. Currently outputting an element as a string decodes these entities, so for example an `&amp;` appears as an `&` and an `&nbsp;` appears as a ` ` (space).

Here's a simple to test to demonstrate desired behaviour:

```
const {document} = parseHTML('<!DOCTYPE html><html id="html" class="live"><!--&lt;hello&gt;-->&lt;hello&gt;</html>');

document.documentElement.innerHTML = '<div>Foo&nbsp;&nbsp;&nbsp;Bar</div>'

// Assertion fails currently (desired behaviour)
assert(document.documentElement.toString(), '<div>Foo&nbsp;&nbsp;&nbsp;Bar</div>')

// Assertion passes currently
assert(document.documentElement.toString(), '<div>Foo    Bar</div>')
```

WebReflection:
This MR does the following:

  * it keeps parsing of entities decoded
  * it escapes only `<`, `>`, and `&` in text and comments, plus `160` char as `&nbsp;`

This MR should close https://github.com/WebReflection/linkedom/pull/64 too.

### DIFF
--- a/cjs/interface/comment.js
+++ b/cjs/interface/comment.js
@@ -1,8 +1,7 @@
 'use strict';
-const {escape} = require('html-escaper');
-
 const {COMMENT_NODE} = require('../shared/constants.js');
 const {VALUE} = require('../shared/symbols.js');
+const {escape} = require('../shared/text-escaper.js');
 
 const {CharacterData} = require('./character-data.js');
 
--- a/cjs/interface/element.js
+++ b/cjs/interface/element.js
@@ -1,8 +1,6 @@
 'use strict';
 // https://dom.spec.whatwg.org/#interface-element
 
-const {unescape} = require('html-escaper');
-
 const {
   ATTRIBUTE_NODE,
   COMMENT_NODE,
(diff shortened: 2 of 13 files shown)


WebReflection:
This MR should close https://github.com/WebReflection/linkedom/pull/64 too.

One thing I am trying to wrap my head around, is that the text escaper rightly parses only `<`, `>`, and `&`, as specs suggest, but I'm not sure what `<div>&amp;amp;</div>` would do ... likely it'd be converted into `&amp;` but then it should go back to `&amp;amp;` if that was the original input, so I need to improve somehow "something" ... either flipping the logic around or something else.

WebReflection:
OK, latest change should address that specific `&` use case.

henricook:
I'm sorry this took me so long, it appears this works for us up to version 0.9.3 and then something breaks again. This is fine for us right now but we will eventually try to upgrade, i'll be in touch and thanks again!

Issue #73, opened 2021-07-01
StyleSheet implementation


Kal-Aster:
Hi, thanks for this library.

Is [StyleSheet](https://developer.mozilla.org/en-US/docs/Web/API/StyleSheet) (`HTMLStyleElement.sheet` property) available?
It always returns me `undefined` and I can't find the implementation in the source code.

WebReflection:
<iframe src="https://giphy.com/embed/in6mnJNYjGKpq" width="480" height="320" frameBorder="0" class="giphy-embed" allowFullScreen></iframe><p><a href="https://giphy.com/gifs/reaction-in6mnJNYjGKpq">via GIPHY</a></p>

that's a big one ... no time (nor needs) here, but MR welcome 👋

WebReflection:
**actually** ... there is a module for that: https://github.com/NV/CSSOM/blob/master/lib/index.js

*maybe* I could `import` that and add it as setter? not sure how it'll work though, but I am not planning to re-write that part.

Kal-Aster:
I think I can help doing it. I just need some time to wrap my head around the source code, it's quite different from how I use to program.

If you could point me where this patch should go, more or less, that would be a great start!

WebReflection:
this is how I'd go:

```js
import {CSSStyleSheet} from 'cssom';

import {registerHTMLClass} from '../shared/register-html-class.js';

import {TextElement} from './text-element.js';

const tagName = 'style';

/**
 * @implements globalThis.HTMLStyleElement
 */
class HTMLStyleElement extends TextElement {
  constructor(ownerDocument, localName = tagName) {
    super(ownerDocument, localName);
  }
  get sheet() {
    return new CSSStyleSheet();
  }
}

registerHTMLClass(tagName, HTMLStyleElement);

export {HTMLStyleElement};
```

Kal-Aster:
Ok, thanks

However it is not problematic, it's just a project structure different from those of mine, so I have to explore a bit to orient myself. I read the "how to contribute" and I think it's explains the project well enough

Kal-Aster:
I just clone the repo and run `npm test` and it gives me
```cmd
> linkedom@0.9.7 test
> eslint esm/ && npm run hello && node test/benchmark/linkedom.js --w3c --mutation-observer --custom-elements && c8 node test/index.js && c8 report -r html       

[...]

PR #74, opened 2021-07-01
Add `sheet` property to `HTMLStyleElement`

Kal-Aster:
Solves #73


cjs/html/style-element.js
```diff
@@ -1,16 +1,51 @@
 'use strict';
+const {parse} = require('cssom');
+
 const {registerHTMLClass} = require('../shared/register-html-class.js');
 
 const {TextElement} = require('./text-element.js');
 
 const tagName = 'style';
 
+const SHEET = Symbol('sheet');
```
WebReflection, line 10: do you mind moving this into https://github.com/WebReflection/linkedom/blob/main/esm/shared/symbols.js ?

cjs/html/style-element.js
```diff
@@ -1,16 +1,51 @@
...
  */
 class HTMLStyleElement extends TextElement {
   constructor(ownerDocument, localName = tagName) {
     super(ownerDocument, localName);
+    this[SHEET] = null;
+  }
+
+  get sheet() {
+    const sheet = this[SHEET];
+    if (sheet != null) {
```
WebReflection, line 23: strict equality please, there's no way this can be *not* `null` at any time: it's either exactly `null` or whatever `parse(...)` returns, which I *hope* is never `null` or `undefined`! `!== null` might be better, thanks.

cjs/html/style-element.js
WebReflection: P.S. apologies I've commented on the `cjs` file, I meant it on the `esm` one.

cjs/html/style-element.js
WebReflection: P.S. apologies I've commented on the `cjs` file, I meant it on the `esm` one.

cjs/html/style-element.js
Kal-Aster: I put it there because I saw the same structure in "template-element" with CONTENT symbol.
Do you want me to move that Symbol too? And maybe do a quick check for eventual others not shared symbols?

WebReflection: Nope, thanks though, that’s been me lazy then, but moving those would be unrelated to this MR, I’ll take care of that

WebReflection: I've added an issue for that https://github.com/WebReflection/linkedom/issues/77

WebReflection:
coverage is not happy ... but I think you could put some  *ignore* around those accessors, as nothing is being tested there, but I don't see side effects ... up to you: ideally you should test some expectations per each accessor, but as all accessors are covered, you can just skip those checks for the time being.

WebReflection:
P.S. ignores are like `/* c8 ignore start */` and `/* c8 ignore stop */` around methods/accessors that don't really need coverage, like yours

Kal-Aster:
I did it \*w*

Kal-Aster:
I did it \*w*

WebReflection:
@Kal-Aster congrats, you're online now, `0.10` it is 🥳

PR #80, opened 2021-07-03
Adds an esm bundle

### LINKED ISSUE #79: Deno support
Hello, I noticed that there is the `rollup/es.config.js` which (I think?) is unused. So I modified it and built an ESM bundle of this library and published it as a fork which you can see here: https://unpkg.com/@matthewp/linkedom@0.10.0/es.js. This works in Deno.

First, are you interested in a build that works with Deno? If not that's fine, I can maintain my fork. Otherwise I'd be happy to contribute here.

My experiment is missing a few things:

* The canvas CJS stuff didn't bundle. I didn't look into why, I just removed it. Could possibly be fixed.
* I removed the perf_hooks import, but I think that can be polyfilled in a bundle.
* I didn't attempt to run any tests in Deno so how *much* works is still an unknown 😀, but a smoke test works so I'm happy.

matthewp:
This changes the rollup config so that it can build an esm bundle. This
excludes the canvas module which gets shimmed to a noop.

You can run with:

```
npx rollup -c rollup/es.config.j
```

I didn't add a build script or publish hook as I wasn't sure how you
felt about those, figured you'd adjust that to your liking anyways.

Adds a test that this works in a web worker. I've tested in Deno as well
and it works there too.

Cheers!

Closes #79


WebReflection:
Thank you! I’m off my laptop watching Europeans but I’ll have a look ASAP and publish it when I can. Appreciated ❤️

matthewp:
No rush at all, enjoy your weekend! Thanks for linkedom.

WebReflection:
Btw, I don’t deno much, is there any particular step I should do to make it available in there too? Is this MR covering it already? Thanks

.gitignore
```diff
@@ -1,4 +1,5 @@
 .DS_Store
 .nyc_output
 coverage/
-node_modules/
\ No newline at end of file
+node_modules/
+es.js
```
WebReflection, line 5: any reason to ignore this file? it could be used even in GitHub pages to test some stuff live. Maybe `worker.js` or `web.js` would be a better name I could add in the exports dictionary too. What do you think?

test/assert-es.js
```diff
@@ -0,0 +1,15 @@
+export default {
+for: Class => {
+    console.log('\x1b[1m', Class.padEnd(60, ' '), '\x1b[0m');
```
WebReflection, line 3: If this is going to be tested on Web, I'd rather go for:

```js
console.log('%c' + Class, 'font-weight:bold');
```

**and** I would remove the `process.exit(1)` as that would fail either ways.

rollup/es.config.js
```diff
@@ -1,16 +1,46 @@
...
+      switch(id) {
+        case 'shim:perf_hooks': {
+          return `
+            export const performance = globalThis.performance;
+          `;
+        }
+        case 'shim:canvas': {
+          return `
+            export default {
+              createCanvas: () => {}
```
WebReflection, line 39: maybe we can use the callback here, so that it won't fully break when/if used.

```js
        return `
          class Canvas {
              constructor(width, height) {
                this.width = width;
                this.height = height;
              }
              getContext() { return null; }
              toDataURL() { return ''; }
            }
            export default (width, height) => new Canvas(width, height);
          `;
```

This is basically the same fallback used in `canvas.cjs`.

.gitignore
matthewp: No reason, I was unsure how you liked to do this. I agree that `es.js` isn't the best name. Between those I think I like `worker.js` slightly better. Let me know what you want and I'll update it.

test/assert-es.js
matthewp: Whoops, yep I'll update this, I pretty much just copied and pasted without really looking at the code.

matthewp:
I've updated based on the review feedback. Went with `worker.js` and added to the export map. Let me know if you want to change that.

----

> Btw, I don’t deno much, is there any particular step I should do to make it available in there too? Is this MR covering it already? Thanks

Since Deno uses URLs for imports you can publish to npm and use in Deno using unpkg or skypack or cdnjs, those will all work without having to do anything. If you want it to feel more official you can go to https://deno.land/x and go through the steps of setting up being published there. It works based on tags on GitHub, so I think you won't have to do anything other than tag your releases (which you already do) once you have it set up.

test/assert-es.js
```diff
@@ -0,0 +1,15 @@
...
+for: Class => {
+    console.log('%c' + Class, 'font-weight:bold');
+    return (expression, expected, message = '') => {
+      if (!Object.is(expression, expected)) {
+        if (message)
+          console.error('', message);
+        console.error('', 'Expected:', String(expected));
+        console.error('', 'Got instead:', String(expression));
+        console.error('');
+        process.exit(1);
```
WebReflection, line 11: nope, I think this still need to go away from here

WebReflection:
@matthewp thanks a lot! everything looks fine except the `process.exit(1)` which breaks on Web and shouldn't exist in ESM, but I don't know if deno has it even when it's not required ... that's the only thing, then I'll merge and publish 👍

test/assert-es.js
WebReflection: Actually ... you know what? I'll drop this, as I need to go soon, no need to hang in here

matthewp: Ah yeah I missed that, my bad.

WebReflection:
@matthewp it's up and running as `0.11` 🥳

matthewp:
I can confirm it's working! Thanks!

deckchairlabs:
@matthewp How do you get the Typescript definitions working in Deno?

matthewp:
@deckchairlabs I'm not sure, I don't use TypeScript.

PR #84, opened 2021-07-18
Extend EventTarget to support event bubbling

mikemadest:
This solves: https://github.com/ungap/event-target/issues/8
As discussed there, EventTarget is extended to add bubbling and parent node knowledge only in the DOM context.


cjs/interface/event-target.js
```diff
@@ -6,4 +6,22 @@ const EventTarget = (m => /* c8 ignore start */ m.__esModule ? m.default : m /*
 /**
  * @implements globalThis.EventTarget
  */
-exports.EventTarget = EventTarget;
+class DOMEventTarget extends EventTarget {
+  getParent() {
```
mikemadest, line 10: specs mention this:

> Each EventTarget object also has an associated get the parent algorithm, which takes an event event, and returns an EventTarget object. Unless specified otherwise it returns null.

So here we don't need to do anything with the event and although I've seen code taking this literally by creating `getTheParent` I just chose `getParent`, it could also be `_getTheParent` or something like that.
This return null here and `Node` returns `parentNode` if it makes sense.

cjs/interface/event.js
```diff
@@ -20,6 +20,8 @@ const GlobalEvent = typeof Event === 'function' ?
     constructor(type, eventInitDict = {}) {
       this.type = type;
       this.bubbles = !!eventInitDict.bubbles;
+      this._stopPropagationFlag = false;
+      this._stopImmediatePropagationFlag = false;
```
mikemadest, line 24: `_stopImmediatePropagationFlag` will be needed, though EventTarget from https://github.com/ungap/event-target don't currently support that feature I thought it was still interesting to add it.

esm/interface/event-target.js
```diff
@@ -5,4 +5,22 @@ import EventTarget from '@ungap/event-target';
...
  * @implements globalThis.EventTarget
  */
-export {EventTarget};
+class DOMEventTarget extends EventTarget {
+  getParent() {
+    return null;
+  }
+
+  dispatchEvent(event) {
+    super.dispatchEvent(event);
```
WebReflection, line 14: I think the correct way here is `const dispatched = super.dispatchEvent(event);` so that you do anything else *only* if `dispatched` is `true`, and you return `dispatched` at the end.

esm/interface/event-target.js
```diff
@@ -5,4 +5,22 @@ import EventTarget from '@ungap/event-target';
 /**
  * @implements globalThis.EventTarget
  */
-export {EventTarget};
+class DOMEventTarget extends EventTarget {
+  getParent() {
```
WebReflection, line 9: I'd rather go for `_getParent()` because having a "*global method*" might be future hostile.

esm/interface/node.js
```diff
@@ -141,6 +141,10 @@ export class Node extends EventTarget {
     return false;
   }
 
+  getParent() {
```
WebReflection, line 144: Same as before, let's use `_getParent()` here, as implementation detail.

WebReflection:
Couple of changes needed, and I'd love to have `stopImmediatePropagation()` tested too, otherwise, awesome job!

mikemadest:
> Couple of changes needed, and I'd love to have `stopImmediatePropagation()` tested too, otherwise, awesome job!

Thanks!
I can't really test the full use case without updating "ungap/event-target" (I could do a PR for that later if that's working for you), but I'll add more tests to cover `stopImmediatePropagation` here.

WebReflection:
> > Couple of changes needed, and I'd love to have `stopImmediatePropagation()` tested too, otherwise, awesome job!
> 
> Thanks!
> I can't really test the full use case without updating "ungap/event-target" (I could do a PR for that later if that's working for you), but I'll add more tests to cover `stopImmediatePropagation` here.

Oh ... I see. Sorry, busy days in here ... I'll try to update/change that later on but if you are, somehow, in a rush, please go ahead with the PR (ungap is a bit weird to work with though, no modern JS and stuff) ... as you prefer.

esm/interface/node.js
```diff
@@ -141,7 +141,7 @@ export class Node extends EventTarget {
     return false;
   }
 
-  getParent() {
+  _getParent() {
```
WebReflection, line 144: Last question, sorry ... could we flag this somehow as `@private` ? I am not sure I want `_` prefixed methods in the TS definition file, although I don't use/know TS much, so if it's pointless, feel free to ignore.

mikemadest:
> Oh ... I see. Sorry, busy days in here ... I'll try to update/change that later on but if you are, somehow, in a rush, please go ahead with the PR (ungap is a bit weird to work with though, no modern JS and stuff) ... as you prefer.

No worries, I didn't mean to rush you :)
I can't say I am in a particular hurry, but I can understand you would be busy and I don't mind at all doing a PR there (especially since I already pulled the code and played with it... so it's mostly ready).
To be honest the only thing I was confused about were the unit tests (in ungap) and for my local tests I used your `assert` there (but I'll refactor that to follow existing code).

esm/interface/node.js
mikemadest: I'm not sure about TS but happy to add that 👍

mikemadest: Is it private or protected? `EventTarget.dispatchEvent` is referring to `EventTarget. _getParent` but we need Node to overwrite `_getParent` and I'm a bit confused as how TS will behave in that case.

After checking TS docs and the net a bit I found this:
https://stackoverflow.com/questions/64083601/overriding-private-methods-in-typescript
which seem to confirm the use of `protected` if that's ok for you?

test/interface/event-target.js
```diff
@@ -0,0 +1,94 @@
...
+  'click',
+  (event) => {
+    event.stopImmediatePropagation();
+    callCount++;
+  },
+  {
+    once: true,
+  },
+);
+containerTarget.addEventListener('click', basicHandler, { once: true });
```
mikemadest, line 90: Testing `stopImmediatePropagation` there.
Once EventTarget support that feature I'll add more event handlers to confirm the full behaviour.

esm/interface/node.js
WebReflection: whatever works, hoping it won't end up in the TS definition 👍

cjs/interface/event-target.js
```diff
@@ -6,4 +6,33 @@ const EventTarget = (m => /* c8 ignore start */ m.__esModule ? m.default : m /*
...
+    if (dispatched && event.bubbles && !event.cancelBubble) {
+      const parent = this._getParent();
+      if (parent && parent.dispatchEvent) {
+        const options = {
+          bubbles: event.bubbles,
+          cancelable: event.cancelable,
+          composed: event.composed,
+        };
+        // in Node 16.5 the same event can't be used for another dispatch
+        return parent.dispatchEvent(new event.constructor(event.type, options));
```
mikemadest, line 31: Changed to cope with Node 16.5 restrictions. I don't know if there is a better way to do that

cjs/interface/event.js
```diff
@@ -36,8 +38,14 @@ const GlobalEvent = typeof Event === 'function' ?
...
-    stopPropagation() {}
-    stopImmediatePropagation() {}
+    stopPropagation() {
+      this.cancelBubble = true;
+    }
+    
+    stopImmediatePropagation() {
+      this.stopPropagation();
+      this._stopImmediatePropagationFlag = true;
+    }
```
mikemadest, line 48: @WebReflection Node 16.5 has issues here:
`stopImmediatePropagation` should set both "stop propagation" and "stop immediate propagation" flags, but Node don't do that and only set their inner [kStop] value (specs https://dom.spec.whatwg.org/#dom-event-stopimmediatepropagation VS implementation of Event in https://github.com/nodejs/node/blob/v16.5.0/lib/internal/event_target.js).

- We have no way of knowing [kStop] as currently they don't have getter for it.

So with that I don't see any way for DOMEventTarget to prevent parent nodes listeners to be called,
unless maybe by extending node's Event with something like `DOMEvent`?

test/interface/event-target.js
```diff
@@ -0,0 +1,101 @@
...
+// https://dom.spec.whatwg.org/#dom-event-stopimmediatepropagation
+// but Node don't do that - will check if that's a bug or expected for them
+const isNode16 = Event._stopImmediatePropagationFlag !== false;
+buttonTarget.addEventListener(
+  'click',
+  (event) => {
+    event.stopImmediatePropagation();
+    if (isNode16) {
+      event.stopPropagation();
+    }
```
mikemadest, line 90: I'm not too happy with that but that's until we decide what to do about Node 16.5 not properly implementing `stopImmediatePropagation`.

cjs/interface/event.js
WebReflection: we are extending the base class though, why wouldn't this work in our `DOMEventTarget` ?

```js
stopImmediatePropagation() {
  super.stopPropagation();
  super.stopImmediatePropagation();
}
```

WebReflection: yes ... but ... doesn't that work on Event? not sure I am following

cjs/interface/event.js
```diff
@@ -43,11 +43,23 @@ const GlobalEvent = typeof Event === 'function' ?
...
+    // https://dom.spec.whatwg.org/#dom-event-stopimmediatepropagation
+    // but Node don't do that so for now we extend it
+    stopImmediatePropagation() {
+      super.stopPropagation();
+      super.stopImmediatePropagation();
+    }
+  }
+  
+
+exports.Event = DOMEvent;
```
WebReflection, line 63: Yeah ... this is what I meant ... it should be fine, but it might break if `stopImmediatePropagation` is not implemented in `GlobalEvent` ... whatever that is ... maybe a try/catch call to `super.stopImmediatePropagation();` could fix that, but it looks a bit dirty.

esm/interface/event.js
```diff
@@ -42,11 +42,23 @@ const GlobalEvent = typeof Event === 'function' ?
...
     
     stopImmediatePropagation() {
-      this.stopPropagation();
       this._stopImmediatePropagationFlag = true;
     }
   };
 
-export {GlobalEvent as Event};
+
+
```
WebReflection, line 50: I think we still need this before:

```
/**
 * @implements globalThis.Event
 */
```

cjs/interface/event.js
mikemadest: So with node 12 and 14 everything works because it relies fully on linkedom implementations.

Node 15+ have Event in that case we use Node's versions which come with an incomplete `stopImmediatePropagation` (they don't have bubbling so for them it probably makes sense).

In the same way we had to extend EventTarget with DOMEventTarget I think it makes sense to extend Event too,
to add what is needed.

This is what I did in the last update, which ensure it works for all versions of node.

cjs/interface/event.js
mikemadest: That's a fair point. I would expect the environment to be Node. Is there any way `Event` could be defined without implementing `stopImmediatePropagation` which is part of the Event spec?

I can add this to be safe:
```javascript
      if (typeof super.stopImmediatePropagation === 'function')
        super.stopImmediatePropagation();
```

mikemadest: Added to be on the safe side

cjs/interface/event.js
WebReflection: > they don't have bubbling so for them it probably makes sense

well ... not really, `stopImmediatePropagation` means don't execute any other listener, bubbling a part ... but I don't care patching the world here, just trying not to break what was working before ;-)

cjs/interface/event-target.js
WebReflection: FWIWI the same happens in the DOM.

mikemadest: Interesting, thanks!

WebReflection:
So ... what are we missing here? is that still ungap for the event target? Thanks!

mikemadest:
> So ... what are we missing here? is that still ungap for the event target? Thanks!

yes, I can make a PR for that next, it should be way simpler.

test/interface/event-target.js
WebReflection: sorry it took a while, but I've published `@ungap/event-target` and now everything "*should*" 🤞 come together

mikemadest: I saw that, thanks! I thought it was quite fast tbh. I'll add more tests probably tomorrow.

As a side note I created a PR in node js about `stopImmediatePropagation`:
https://github.com/nodejs/node/pull/39463

WebReflection: Nice one!

mikemadest: test updated, now add 3 listeners to the button and one to the parent, check that `stopImmediatePropagation` stop running listeners at the right moment.

WebReflection:
@mikemadest this looks good to me, but I don't understand why there are unrelated changes in this MR ... by any chance you can remove those and pull rebase from main instead?

```sh
git pull --rebase origin main
```

I'd rather keep changes per PR/MR, thanks. Once you do that, I'll squash merge and publish.

mikemadest:
> @mikemadest this looks good to me, but I don't understand why there are unrelated changes in this MR ... by any chance you can remove those and pull rebase from main instead?
> 
> ```shell
> git pull --rebase origin main
> ```
> 
> I'd rather keep changes per PR/MR, thanks. Once you do that, I'll squash merge and publish.

Sorry I'm not sure what happened there, will cleanup

mikemadest:
> @mikemadest this looks good to me, but I don't understand why there are unrelated changes in this MR ... by any chance you can remove those and pull rebase from main instead?
> 
> ```shell
> git pull --rebase origin main
> ```
> 
> I'd rather keep changes per PR/MR, thanks. Once you do that, I'll squash merge and publish.

ok so I was just behind main because of other merge, this looks all good now

WebReflection:
`v0.11.1` is up and running 🍻

Thanks a lot for your help ❤️

mikemadest:
> `v0.11.1` is up and running 🍻
> 
> Thanks a lot for your help ❤️

thank you for bearing with me :)

PR #85, opened 2021-07-19
Adding NamedNodeMap to global export

mikemadest:
The reason for this PR is this error when running Axe tests in a React app, 
using Jest with a linkedom environment:

image

And this is the axe test involved:
```javascript
    function getNodeAttributes(node) {
      if (node.attributes instanceof window.NamedNodeMap) {
        return node.attributes;
      }
      return node.cloneNode(false).attributes;
    }
```

What happen is that `NamedNodeMap`, despite being in linkedom interface, wasn't available in `window`, which is what this PR intend to change.


test/interface/named-node-map.js
```diff
@@ -2,10 +2,10 @@ const assert = require('../assert.js').for('NamedNodeMap');
 
 const {parseHTML} = global[Symbol.for('linkedom')];
 
-const {document} = parseHTML('<html><div tmp>abc</div></html>');
+const {document, NamedNodeMap} = parseHTML('<html><div tmp>abc</div></html>');
 
 let node = document.documentElement.firstElementChild;
-
+assert(typeof NamedNodeMap !== 'undefined', true, 'NamedNodeMap undefined in global export');
```
mikemadest, line 8: My intention was to add a unit test related to the change here. I couldn't find similar checks for other exported objects but felt bad not adding something...

esm/interface/document.js
```diff
@@ -30,6 +30,7 @@ import {InputEvent} from './input-event.js';
 import {ImageClass} from './image.js';
 import {MutationObserverClass} from './mutation-observer.js';
 import {NodeList} from './node-list.js';
+import {NamedNodeMap} from './named-node-map.js';
```
WebReflection, line 33: Do you mind moving this above the previous line? Just to preserve the alphabetical order, thanks!

WebReflection:
Thanks, beside one line to move, this looks good to me 👍

esm/interface/document.js
mikemadest: Of course, thanks for your feedback 👍

mikemadest: Done

PR #89, opened 2021-08-02
Event bubbling keep target

mikemadest:
The first implementation of bubbling with `EventTarget` was missing important parts:
- document get-the-parent algorithm wasn't pointing to `window`
- `event.target` should be where the initial dispatch happened (no change for `event.currentTarget`, which is where the listener is attached).

Because NodeJS and @ungap/event-target kept `target` and `currentTarget` in sync they are not relied on anymore since this would prevent solving this issue.

Other change: 
- `eventPhase` now is initially at NONE, switch to `CAPTURING_PHASE` when an event is dispatched, then `AT_TARGET` when on its target and finally `BUBBLING_PHASE` after going through target:
https://whatpr.org/dom/699.html#dom-event-eventphase
- event now has `composedPath()` method (though a very simple version of it). That was done because it's closer to how it should be while not adding too much code.

This was tested on Node v12.22.4, v14.17.4, v16.5.0 and v16.6.0.


cjs/interface/document.js
```diff
@@ -159,6 +159,13 @@ class Document extends NonElementParentNode {
 
   get isConnected() { return true; }
 
+  /**
+   * @protected
+   */
+   _getParent() {
+    return this[EVENT_TARGET];
```
mikemadest, line 166: returning `this.defaultView` wasn't working as expected and since this is used only for bubbling that was updated to work in that context.

cjs/interface/event-target.js
```diff
@@ -1,12 +1,45 @@
 'use strict';
 // https://dom.spec.whatwg.org/#interface-eventtarget
 
-const EventTarget = (m => /* c8 ignore start */ m.__esModule ? m.default : m /* c8 ignore stop */)(require('@ungap/event-target'));
+const wm = new WeakMap();
```
mikemadest, line 4: instead of extending Node or ungap EventTarget all features are now here.
In the end I think this is better as it allow for easier changes (one repo) and doesn't let the code exposed to eventual Node changes that doesn't support needed features anyway.

cjs/interface/event-target.js
```diff
@@ -15,24 +48,37 @@ class DOMEventTarget extends EventTarget {
...
+    if (listeners.some(info => info.listener === listener)) {
+      return;
+    }
+    listeners.push({target: this, listener, options});
+  }
+
+  removeEventListener(type, listener) {
+    const secret = wm.get(this);
+    const listeners = secret[type] || (secret[type] = []);
+    secret[type] = listeners.filter(info => info.listener !== listener);
```
mikemadest, line 63: not sure if we want to do that or keep the `splice` to actually update the initial array. I chose to update it for the sake of the discussion.

cjs/interface/event-target.js
```diff
@@ -15,24 +48,37 @@ class DOMEventTarget extends EventTarget {
...
+    while (node) {
+      if (node.dispatchEvent)
+        event._path.push({currentTarget: node, target: this});
+      node = event.bubbles && node._getParent && node._getParent();
     }
-    return dispatched;
+    event._path.some(invokeListeners, event);
+    event._path = [];
+    event.eventPhase = event.NONE;
+    return !event.defaultPrevented;
```
mikemadest, line 79: initially it was just returning true, but doc says 

> Return false if event’s canceled flag is set, and true otherwise.

WebReflection:
this is a "big one" ... I'll try to have a look soon, but in general, even if I appreciate you commenting on your own MR before me asking anything, it makes code review instantly more problematic as comments break the review flow 😅

so, please give me some time to review this fresh, then read through your comments, then eventually comment or approve, thank you!

mikemadest:
> this is a "big one" ... I'll try to have a look soon, but in general, even if I appreciate you commenting on your own MR before me asking anything, it makes code review instantly more problematic as comments break the review flow 😅
> 
> so, please give me some time to review this fresh, then read through your comments, then eventually comment or approve, thank you!

My bad, this may be a bad habit as I try to explain why I did things, probably a bit too much. I'll avoid doing that in the future 👍

WebReflection:
> this may be a bad habit as I try to explain why I did things

I appreciate the intent, it's just impractical to my eyes, so it's likely my fault, as I want to see the code, and ask questions once I've seen it all, usually not the other way around.

I will take some time to get rid of "the comments noise" and go through these changes though, just please be patient 👍

esm/interface/event-target.js
```diff
@@ -1,11 +1,44 @@
...
+}
 
 /**
  * @implements globalThis.EventTarget
  */
-class DOMEventTarget extends EventTarget {
+class DOMEventTarget {
+
+  constructor() {
+    wm.set(this, Object.create(null));
```
WebReflection, line 40: we need a `null` Dictionary class here, possibly in a file a part.

```js
class Dictionary {}
Dictionary.prototype = Object.create(null);
```

Alternatively, I'd like to use `new Map` and related map utilities to handle `secret`. These are performance improvements, so that if none of this is your cup of tea, at least use `const {create} = Object` so no lookup and no patches to `Object.create` can ever affect the result.

WebReflection:
CI says "*This pull request is still a work in progress*" ... so I can't merge it, beside the mentioned changes on `secret`. Please flag it as ready whenever you feel like.

esm/interface/event-target.js
mikemadest: I find that suggestion interesting and will update the PR 👍

mikemadest: I couldn't follow the suggestion exactly as Dictionary.prototype was read only and that code was showing an error,
but I tried something else. There is also a change to use Map, not sure if the 1:1 replacement was what you had in mind and I didn't want to stray too far from the original code.

mikemadest:
Updated with
- new Dictionary in "shared",
-renamed `wm` to `eventTargetMap` (I could have used `m` but it felt a bit short and not self documenting enough).

WebReflection:
Please rebase to avoid conflicts. I needed to move away from Travis CI and the update changed some TS definition. Thanks for your patience, apologies for the delay here ... it's been very busy times.

esm/shared/dictionary.js
```diff
@@ -0,0 +1,7 @@
+const {create} = Object;
+
+export class Dictionary {
+  constructor() {
+    this.prototype = create(null);
```
WebReflection, line 5: nope ... this is not what I meant ... `Dictionary.prototype = create(null);` is what I was after, but on a closer look, we might want to get rid of this class completely and use a `Map` instead for listeners, so we don't clash, ever, and we have Map methods to deal with listeners in a natural way.

mikemadest:
> Please rebase to avoid conflicts. I needed to move away from Travis CI and the update changed some TS definition. Thanks for your patience, apologies for the delay here ... it's been very busy times.

No worries I was busy too on another project delivery :)
I'll rebase and check that, thanks!

esm/shared/dictionary.js
mikemadest: I tried your initial suggestion first but that wasn't working and that version was the only way I could find. 
Will update with Map!

mikemadest:
In the end I wasn't sure what to do and went back to the original code, 
with just the ask about `const {create} = object;`.

Notes:
I tried a few things but wasn't happy with any.
NodeJS use symbols and a modified Map (SafeMap), while JSDOM use an EventTarget property to store listeners initialised with Object.create(null).

I tried something like this (but rejected that as not good enough, posting it for the sake of discussion):
```javascript
const kListeners = Symbol('kListeners');
```

```javascript
  constructor() {
    this[kListeners] = new Map();
  }

  addEventListener(type, listener, options) {
    const listeners = this[kListeners].get(type) || [];
    if (listeners.some(info => info.listener === listener)) {
      return;
    }
    listeners.push({target: this, listener, options});
    this[kListeners].set(type, listeners);
  }

  removeEventListener(type, listener) {
    const listeners = this[kListeners].get(type) || [];
    this[kListeners].set(type, listeners.filter(info => info.listener !== listener));
  }
```

I think listeners should be a property of EventTarget in some way.
But that change doesn't need to be part of this PR and I don't want to waste your time trying different changes on that.

cjs/interface/event-target.js
```diff
@@ -1,12 +1,46 @@
 'use strict';
 // https://dom.spec.whatwg.org/#interface-eventtarget
 
-const EventTarget = (m => /* c8 ignore start */ m.__esModule ? m.default : m /* c8 ignore stop */)(require('@ungap/event-target'));
+const eventTargetMap = new Map();
```
WebReflection, line 4: looks like this should  be a `WeakMap` instead.

WebReflection:
@mikemadest for discussion sake, the relation `this` -> *listeners* -> *options* is what we're after, and those are:
  * a `WeakMap`, or a private class field, that holds a `Map`
  * such `Map` stores maps by type/name
  * such stored `Map` uses *listeners* as keys, and *options* as value

```js
constructor() {
  wm.set(this, new Map);
}
addEventListener(type, listener, options) {
  const map = wm.get(this);
  if (!map.has(type))
    map.set(type, new Map);
  map.get(type).set(listener, options);
}
removeEeventListener(type, listener) {
  const map = wm.get(this);
  if (map.has(type))
    map.get(type).delete(listener);
}
dispatchEvent(event) {
  const map = wm.get(this);
  if (map.has(event.type)) {
    const listeners = map.get(event.type);
    for (const [listener, options] of listeners) {
      dispatch({target: this, listener, options});
      if (options.once)
        listeners.delete(listener);
    }
  }
}
```

I hope this example simplifies what I meant by using a map. In terms of definition:

  * `Map<Function|Object, any?>` for the listener/options
  * `Map<string, Map>` for the type/listeners
  * `WeakMap<Node, Map>`

The relation ensures with ease no duplicated listeners are ever set, it doesn't need splice, slice, filter, some, and it should be pretty fast too ... what do you think?

mikemadest:
> @mikemadest for discussion sake, the relation `this` -> _listeners_ -> _options_ is what we're after, and those are:
> 
> * a `WeakMap`, or a private class field, that holds a `Map`
> * such `Map` stores maps by type/name
> * such stored `Map` uses _listeners_ as keys, and _options_ as value
> 
> I hope this example simplifies what I meant by using a map. In terms of definition:
> 
> * `Map<Function|Object, any?>` for the listener/options
> * `Map<string, Map>` for the type/listeners
> * `WeakMap<Node, Map>`
> 
> The relation ensures with ease no duplicated listeners are ever set, it doesn't need splice, slice, filter, some, and it should be pretty fast too ... what do you think?

I updated the code to use that. It seems to work and for performances this seems pretty efficient at least
(sorry for the delay, I didn't have much time lately).

cjs/interface/event-target.js
```diff
@@ -15,24 +54,43 @@ class DOMEventTarget extends EventTarget {
...
+  removeEventListener(type, listener) {
+    const map = wm.get(this);
+    if (map.has(type)) {
+      const listeners = map.get(type);
+      if (listeners.delete(listener) && !listeners.size) {
+        map.delete(type);
+      } else {
+        map.set(type, listeners);
+      }
+      wm.set(this, map);
```
WebReflection, line 74: I am confused ... why are you setting the `this` every time to the same map it had already?

cjs/interface/event-target.js
```diff
@@ -15,24 +54,43 @@ class DOMEventTarget extends EventTarget {
     return null;
   }
 
+  addEventListener(type, listener, options) {
+    const map = wm.get(this);
+    if (!map.has(type)) 
+      map.set(type, new Map);
+    const listeners = map.get(type).set(listener, options);
+    wm.set(this, map.set(type, listeners));
```
WebReflection, line 62: confused in here too ... why are you replacing the map?

cjs/interface/event-target.js
mikemadest: Good point, I had an error because i forgot a genuily usefull `set`, but overdid the "fix" forgetting we store the reference anyway...

WebReflection: to be honest I don't understand neither why you are setting anything on removeEventListener ... if it's not there, don't set it?

mikemadest: All those `set` should be removed now

PR #91, opened 2021-08-09
ShadowRoot should handle innerHTML

mikemadest:
### Why did I do it?

While working with `getRootNode` I came across this example:
[Mozilla's doc on getRootNode](https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode)

This makes use of `ShadowRoot` innerHTML in addition to asserting `getRootNode` behaviour.
As I want to get this example working (and match equivalent results in browser) this innerHTML seemed like a good first part.

### What was done?

- move innerHTML logic to a mixin
- updates to Element and ShadowRoot to use that mixin


esm/mixin/inner-html.js
```diff
@@ -0,0 +1,19 @@
+import {parseFromString} from '../shared/parse-from-string.js';
+import {
+  ignoreCase,
+} from '../shared/utils.js';
+import {
+  CUSTOM_ELEMENTS
+} from '../shared/symbols.js';
+
```
WebReflection, line 8: for consistency sake, please import constants on top and don't use multiple lines if the import fits into 80 chars/column.

esm/mixin/inner-html.js
```diff
@@ -0,0 +1,19 @@
...
+export const getInnerHtml = node => node.childNodes.join('');
```
WebReflection, line 9: Please add return type via JSDoc, as the TS definition now returns `any` instead of `string`.

esm/mixin/inner-html.js
mikemadest: Done.

esm/mixin/inner-html.js
mikemadest: Updated, now show the correct type

mikemadest:
Rebased

Issue #99, opened 2022-01-20
Incorrect types for some `Node` methods


gingerchris:
I'm using linkedom in a typescript project and experiencing some inaccurate typescript errors when using methods like `removeChild` from an instance of `Node`.  The functionality is working as expected, which leads me to believe the error is incorrect.

The error presented is `Expected 0 arguments, but got 1.` which I believe stems from the [declared type for this method here](https://github.com/WebReflection/linkedom/blob/main/types/interface/node.d.ts#L44)

I've put together a minimal reproduction of the issue [here](https://codesandbox.io/s/linkedom-removechild-typescript-error-odmhe)

Also here's a screenshot showing the error:
![Screenshot 2022-01-20 at 12 13 32](https://user-images.githubusercontent.com/1153616/150343361-11b8b34c-4f0a-4819-ac64-09b15e3898d4.png)

Thanks for your work on this project, it's amazing.

WebReflection:
@gingerchris thanks for filing this issue but I am not really a TS and everything TS related is usually provided by the community, so any PR would be more than welcome and yes, definitively I see something ain't right there, and probably there are other methods too not fully aligned ... do you mind proposing those changes? Thank you!

gingerchris:
@WebReflection absolutely, I'll pull a PR together.

WebReflection:
This was closed by https://github.com/WebReflection/linkedom/pull/100

PR #100, opened 2022-01-20
Type definition updates

gingerchris:
### What's the problem?

When using linkedom in a Typescript project, some functionality causes Typescript errors, despite that functionality working.  [Here's one example](https://github.com/WebReflection/linkedom/issues/99), another comes from accessing `innerText` on an element returned from a `querySelector`

### What have I done?

This PR updates type definitions for the methods  `insertBefore`, `appendChild`, `replaceChild`,`removeChild` on `Node`. 
 These updated to more closely match the functionality they provide, but still cause errors from the typescript compiler as they don't exactly match those provided by `globalThis.Node`.  After this change, the errors reported when implementing linkedom in another project go away.

This also updates the `NodeStruct` type to extend the `Element` type.  This enables accessing element properties from elements returned by `querySelector`.

The `NodeStruct` update appears to have confused the typescript compiler as some of the derived types were incorrectly converted to `any`.  To overcome this, I have added explicit return types to some other methods.

### DIFF
--- a/cjs/interface/element.js
+++ b/cjs/interface/element.js
@@ -137,6 +137,9 @@ class Element extends ParentNode {
   // <contentRelated>
   get innerText() { return this.textContent; }
 
+  /**
+   * @returns {String}
+   */
   get textContent() {
     const text = [];
     let {[NEXT]: next, [END]: end} = this;
--- a/cjs/interface/node.js
+++ b/cjs/interface/node.js
@@ -84,10 +84,33 @@ class Node extends EventTarget {
   normalize() {}
   cloneNode() { return null; }
   contains() { return false; }
-  insertBefore() {}
-  appendChild() {}
-  replaceChild() {}
-  removeChild() {}
+  /**
+   * Inserts a node before a reference node as a child of this parent node.
+   * @param {Node} newNode The node to be inserted.
+   * @param {Node} referenceNode The node before which newNode is inserted. If this is null, then newNode is inserted at the end of node's child nodes.
+   * @returns The added child
+   */
+  // eslint-disable-next-line no-unused-vars
+  insertBefore(newNode, referenceNode) { return newNode }
+  /**
+   * Adds a node to the end of the list of children of this node.
+   * @param {Node} child The node to append to the given parent node.
+   * @returns The appended child.
+   */
+  appendChild(child) { return child }
+  /**
+   * Replaces a child node within this node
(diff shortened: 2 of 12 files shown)


gingerchris:
@WebReflection I'm not 100% sure the second commit here is doing the right thing.  Is there sufficient crossover between `NodeStruct` and `Element` for this to make sense?  Or do you think it would it make more sense to update the type returned by `QuerySelector` to be `Element`?

WebReflection:
@gingerchris I think I've just realized this project types are defined via `tsc -p .` as build step, so these changes should likely land in the `esm/*.js` related file, not in the generated types, or these will be replaced each time a new build is made.

That means that `esm/interface/node.js` is where those should be updated, and `NodeStruct` is for `parent-node.js` and definitively is about `Element` because non elements cannot contain other nodes, so I'd be OK with the change, as long as it's done in the right place ;-)

gingerchris:
@WebReflection excellent, thanks for your help.  I will make these updates as soon as I get a chance

gingerchris:
@WebReflection I've updated this PR now with changes to the `/esm/` code that resulted in the correct types being generated.  I included all the compiled files for each commit - not sure if that was the right thing to do?

In the third commit, I've had to add JSDOC definitions for some methods that didn't previously need them in order to avoid some of the generated types changing. Is that ok?

WebReflection:
@gingerchris thanks! I've merged, updated all dependencies, and merge published all changes as 0.13.1 ... let me know if something is not working as expected, as the TSC also got updated and now I have more files for types 🤔

gingerchris:
@WebReflection thanks!  [The demo I put together](https://codesandbox.io/s/linkedom-removechild-typescript-error-odmhe?file=/src/index.ts) is no longer showing that error.

WebReflection:
@gingerchris not sure how's that supposed to work without perf_hooks **but** if it's Workers that you are after remember to import `linkedom/worker` instead, as mentioned already in here: https://github.com/WebReflection/linkedom/issues/101

WebReflection:
btw, I think 0.13.2 should stop showing errors around perf_hooks where this is not defined.

gingerchris:
Ah sorry - I should have explained - I wasn't intending for this code to work.  It was just to illustrate that there are no typescript errors now, whereas on the previous release of `linkedom` there were.

Issue #103, opened 2022-01-31
Support for history and location


thejohnhoffer:
Here's a CodeSandbox that shows parseHTML's document.defaultView doesn't contain `history` or `location`:

https://codesandbox.io/s/linkedom-no-history-o666v?file=/src/index.js

### Possible solutions:

When looking for `location` or `history`, I found [a comment](https://github.com/WebReflection/linkedom/issues/50#issuecomment-812931931) by @stephenh in issue #50:
> does linkedom support `window.location`?

I didn't see any folllow-up on that question, so I've used [remix-run/history][history] to simulate both `location` and `history` in [my own wrapper][wrapper] around `linkedom`.
[wrapper]: https://www.npmjs.com/package/linkedom-history

I'd be happy to work on a pull request to mock `history` and `location` within `linkedom`. Now, I use only one function called `createMemoryHistory` from "remix-run/history". Following linkedom's minimalist philosophy, I'd be happy to extract the `createMemoryHistory` function from "remix-run/history".

[history]: https://github.com/remix-run/history

WebReflection:
Any JavaScript in this project has no meaning … there’s no JS here, if not DOM related. I don’t understand use cases around these requests but again, if it’s to simulate a browser, this is not the goal of this project.

if you would like to provide a lazy accessor for history or others API, as long as it’s lazy and doesn’t change anything else around this project, I’d be more than happy to have a look, but if your goal is to have JS running within linkedom documents, I recommend you go back to JSDOM or use cypress, puppeteer, or others solutions.

LinkeDOM is for SSR, not much else, really

thejohnhoffer:
Hi @WebReflection -- thank you so much for this quick and clarifying response.

> if you would like to provide a lazy accessor for history or others API, as long as it’s lazy and doesn’t change anything else around this project, I’d be more than happy to have a look

Here's the quick pitch: if the user already has `remix-run/history` installed, then the `defaultView` would proxy `history` and `location` requests to a `MemoryHistory` from that library. 

> but if your goal is to have JS running within linkedom documents, I recommend you go back to JSDOM or use cypress, puppeteer, or others solutions.

I appreciate the guidance. With that in mind, it seems suprising how well `linkedom` works with `@testing-library/react`. I'd be interested to look deeper into why this `querySelector` call works if `linkedom` wasn't intended to work this way:

```jsx
import { parseHTML } from "linkedom"; 
import { render } from "@testing-library/react";  
[...]

PR #104, opened 2022-02-03
support window exports with WINDOW symbol

thejohnhoffer:
This closes Issue #103. This allows the `HTMLDocument` constructor to accept a special WINDOW  property that sets up lazy access of new properties for:
 - `defaultView` and `defaultView.window`
 -  such as `location` and `history`, or any arbitrary key.  
 
 If the user happens to provide the [special](https://codesandbox.io/s/window-and-document-0pngb?file=/src/index.ts) `location` key, then:
 - The `document.location` will match the `location` of `defaultView` and `defaultView.window`. 
 - Otherwise, the keys will only apply to `defaultView` and `defaultView.window`.

No existing test changed. Three examples run in a new file, [test/html/global.js](https://github.com/thejohnhoffer/linkedom/blob/fe502e177c5ebcc246e7f0c0c77b07c839976bdc/test/html/global.js).

## DOMParser history Example

```js
const {WINDOW} = global[Symbol.for('linkedom')];

let document = (new DOMParser).parseFromString('', {
  mime: 'text/html',
  [WINDOW]: {
    get history() {
      return {
        get location() { return {pathname: '/'}; }
      };
    }
  }
});

assert(document.defaultView.history.location.pathname, '/');
assert(document.defaultView.location, undefined);
```

##  parseHTML history and location Example

```js
const {window, history, location} = parseHTML(`
<!doctype html>
<html>                                 
  <head><title>hello</title></head>
</html>                                                
`, {                                                                                           
  get history() {                     
    return {
      get location() { return {pathname: '/'}; }      
    };                                              
  },                                            
  get location() { return {pathname: '/'}; }
});         

assert(history.location.pathname, '/'); 
assert(window.location.pathname, '/');                    
assert(location.pathname, '/');
```

## DOMParser mutable location Example

```js
document = (new DOMParser).parseFromString('<!DOCTYPE html><html />', {
  [WINDOW]: {      
    location: {    
      pathname: '/'
    }
  } 
});

assert(document.location.hash, undefined);
assert(document.location.pathname, '/');

document.location = {
  pathname: '/url/file.html',
  hash: '#hash-string'
}

assert(document.defaultView.location.pathname, '/url/file.html');
assert(document.defaultView.location.hash, '#hash-string');
assert(document.location.pathname, '/url/file.html');
assert(document.location.hash, '#hash-string');
```

### DIFF
--- a/esm/dom/parser.js
+++ b/esm/dom/parser.js
@@ -4,6 +4,7 @@ import {parseFromString} from '../shared/parse-from-string.js';
 import {HTMLDocument} from '../html/document.js';
 import {SVGDocument} from '../svg/document.js';
 import {XMLDocument} from '../xml/document.js';
+import {MIME} from '../shared/symbols.js';
 
 /**
  * @implements globalThis.DOMParser
@@ -11,17 +12,25 @@ import {XMLDocument} from '../xml/document.js';
 export class DOMParser {
 
   /** @typedef {{ "text/html": HTMLDocument, "image/svg+xml": SVGDocument, "text/xml": XMLDocument }} MimeToDoc */
+  /** @typedef {{ [x: symbol]: unknown, [MIME] : string }} Config */
   /**
    * @template {keyof MimeToDoc} MIME
    * @param {string} markupLanguage
-   * @param {MIME} mimeType
+   * @param {Config | string} config
    * @returns {MimeToDoc[MIME]}
    */
-  parseFromString(markupLanguage, mimeType) {
+  parseFromString(markupLanguage, config) {
+    const [mimeType, domOptions] = ((c) => {
+      if (typeof c === 'object') {
+        const {[MIME]: mime, ...rest} = c;
+        return [mime || 'text/html', rest]
+      }
+      return [c, {}]
+    })(config);
     let isHTML = false, document;
     if (mimeType === 'text/html') {
       isHTML = true;
-      document = new HTMLDocument;
+      document = new HTMLDocument(domOptions);
     }
     else if (mimeType === 'image/svg+xml')
       document = new SVGDocument;
--- a/esm/html/document.js
+++ b/esm/html/document.js
@@ -6,6 +6,7 @@ import {Document} from '../interface/document.js';
 import {NodeList} from '../interface/node-list.js';
 import {customElements} from '../interface/custom-element-registry.js';
 
+import {WINDOW} from '../shared/symbols.js';
 import {HTMLElement} from './element.js';
 
 const createHTMLElement = (ownerDocument, builtin, localName, options) => {
@@ -29,7 +30,7 @@ const createHTMLElement = (ownerDocument, builtin, localName, options) => {
 /**
  * @implements globalThis.HTMLDocument
  */
-export class HTMLDocument extends Document {
+class _HTMLDocument extends Document {
   constructor() { super('text/html'); }
 
   get all() {
@@ -101,4 +102,54 @@ export class HTMLDocument extends Document {
       element.setAttribute('is', options.is);
     return element;
   }
+}
+
+/** @typedef {{ [WINDOW] : Object }} Context */
+/**
+ * @constructor
+ * @param {Context} context
+ */
+export class HTMLDocument extends _HTMLDocument {
+
+  constructor (context = {}) {
+    super();
+    const useState = (target, name) => {
+      return ((_0, _1={}) => {
+        const g = (name in _1) ? _1 : _0;
+        return {
+          get: () => g[name],
+          set: (v) => (g[name] = v) || true,
+        }
+      })(target, context[WINDOW]);
+    };
+    this[WINDOW] = {
+      set: (target, name, value) => {
+        return useState(target, name).set(value);
+      },
+      get: (target, name) => {
+        if (name === 'window') {
+          return new Proxy(target.window, this[WINDOW]);
+        }
+        return useState(target, name).get();
+      }
+    }
+  }
+
+  get defaultView() {
+    return new Proxy(super.defaultView, this[WINDOW])
+  }
+
+  get location() {
+    return this.defaultView.location;
+  }
+
+  set location(location) {
+    this.defaultView.location = location;
+  }
+
+  get all() { return super.all } 
+  get head() { return super.head } 
+  get body() { return super.body } 
+  get title() { return super.title } 
+  set title(v) { super.title = v } 
 }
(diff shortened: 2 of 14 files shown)


esm/dom/parser.js
```diff
@@ -4,24 +4,33 @@ import {parseFromString} from '../shared/parse-from-string.js';
...
    */
-  parseFromString(markupLanguage, mimeType) {
+  parseFromString(markupLanguage, config) {
+    const [mimeType, domOptions] = ((c) => {
+      if (typeof c === 'object') {
+        const {[MIME]: mime, ...rest} = c;
+        return [mime || 'text/html', rest]
+      }
+      return [c, {}]
+    })(config);
```
thejohnhoffer, line 29: The second argument can now look like: 
- "text/html"
- {[WINDOW]: {}}
- {mime: "text/html", [WINDOW]: {}}
- {[MIME]: "text/html", [WINDOW]: {}}

esm/html/document.js
```diff
@@ -102,3 +103,53 @@ export class HTMLDocument extends Document {
     return element;
   }
 }
+
+/** @typedef {{ [WINDOW] : Object }} Context */
+/**
+ * @constructor
+ * @param {Context} context
+ */
+export class HTMLDocument extends _HTMLDocument {
```
thejohnhoffer, line 112: There is no need to keep _HTMLDocument and HTMLDocument separate, but I thought it would be clearer to understand this PR this way.

esm/html/document.js
```diff
@@ -102,3 +103,53 @@ export class HTMLDocument extends Document {
...
+      })(target, context[WINDOW]);
+    };
+    this[WINDOW] = {
+      set: (target, name, value) => {
+        return useState(target, name).set(value);
+      },
+      get: (target, name) => {
+        if (name === 'window') {
+          return new Proxy(target.window, this[WINDOW]);
+        }
```
thejohnhoffer, line 132: This ensures that `defaultView.window` is proxied as well as `defaultView`.

esm/html/document.js
```diff
@@ -102,3 +103,53 @@ export class HTMLDocument extends Document {
...
+    }
+  }
+
+  get defaultView() {
+    return new Proxy(super.defaultView, this[WINDOW])
+  }
+
+  get location() {
+    return this.defaultView.location;
+  }
```
thejohnhoffer, line 144: A special "location" property is shared between `window` and `document`.

thejohnhoffer (commented):
I'm happy to bike-shed any of these points!

esm/dom/parser.js
WebReflection: what's the reason for `MIME` symbol, if I might ask?

esm/html/document.js
WebReflection: I think any document should accept extras, not just the HTML kind

esm/html/document.js
```diff
@@ -102,3 +103,53 @@ export class HTMLDocument extends Document {
...
+ * @constructor
+ * @param {Context} context
+ */
+export class HTMLDocument extends _HTMLDocument {
+
+  constructor (context = {}) {
+    super();
+    const useState = (target, name) => {
+      return ((_0, _1={}) => {
+        const g = (name in _1) ? _1 : _0;
```
WebReflection, line 118: this easily leaks the Object.prototype namespace to the window ... probably not a real-world concern, yet `hasOwnProperty` might be better

esm/html/document.js
WebReflection: if this is how the standard works I'm OK with it, if not, we need to think about a better approach that is more flexible around shared things

esm/html/document.js
```diff
@@ -102,3 +103,53 @@ export class HTMLDocument extends Document {
...
+
+  set location(location) {
+    this.defaultView.location = location;
+  }
+
+  get all() { return super.all } 
+  get head() { return super.head } 
+  get body() { return super.body } 
+  get title() { return super.title } 
+  set title(v) { super.title = v }
```
WebReflection, line 154: I have no idea why any of these is needed

WebReflection:
thanks! I have a busy day ahead but I'll have a look at some point ... I am not sure all changes are needed, and I had in mind something slightly simpler still similar ... bear with me I'll come back 👋

esm/html/document.js
thejohnhoffer: Okay, I agree that seems like a good idea.

esm/dom/parser.js
thejohnhoffer: The symbols here are not needed. The only benefit now is that the `config` object has only `Symbol` keys. It would be fine to have the string "mime" and the symbol `WINDOW` for keys...

Actually, it would even be okay to have the strings "mime" and "window" as keys. At the `parseFromString` step, there is currently no risk of namespace conflict. The `WINDOW` symbol of "linkedomWindowProxy" only must be used internally [here](https://github.com/WebReflection/linkedom/pull/104/files#diff-c5f236b00784b2abd39c39c877ff6115c540f6cf142112bba3ad49ae7bfda28cR125).

esm/html/document.js
thejohnhoffer: A quick [code sandbox](https://codesandbox.io/s/window-and-document-0pngb?file=/src/index.ts) identifies only "location" as the only non-null shared property that linkedom doesn't yet cover. However, I've recently noticed 89 properties like "onload" that default to `null` in both `document` and `window`.

For completeness, we should check [the HTML standard](https://html.spec.whatwg.org/) whether those 89 properties identify the same objects in both [window](https://html.spec.whatwg.org/multipage/window-object.html#window) and [document](https://html.spec.whatwg.org/multipage/dom.html#document).

esm/html/document.js
thejohnhoffer: Here's [an example line](https://github.com/WebReflection/linkedom/blob/main/esm/cached.js#L125) that causes an issue during build without `get title()` duplicated. The `Object.getOwnPropertyDescriptors` method does not seem to detect properties from the base class.  Apart from merging `HTMLDocument` and `_HTMLDocument`, the solution would be to avoid calling `getOwnPropertyDescriptors` on `HTMLDocument`.

thejohnhoffer:
> thanks! I have a busy day ahead but I'll have a look at some point ... I am not sure all changes are needed, and I had in mind something slightly simpler still similar ... bear with me I'll come back 👋

I'm sure there are ways to simplify. On my own fork, I'll track planned changes [as issues](https://github.com/thejohnhoffer/linkedom/issues). I'll hold off on most of these until I hear back.

WebReflection:
Fixed in a different way and explained in here: https://github.com/WebReflection/linkedom/issues/103#issuecomment-1036101047

PR #107, opened 2022-02-09
fix: this within custom elements constructor is not the same as in method calls

tmaroschik:
During the initialization of a lit-element based custom element, some references to `this` are captured within the constructor. As the element becomes the prototype of an `HTMLElement` via `Object.setPropertyOf` the captured references to `this` are broken.

After spending some time researching how to fix this, I stumbled over a solution in the webcomponentsjs polyfill https://github.com/webcomponents/polyfills/blob/master/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js#L161

This PR is a close adaption of their approach and makes our SSR rendering work with lit-elements as well.

### DIFF
--- a/cjs/html/element.js
+++ b/cjs/html/element.js
@@ -4,7 +4,7 @@ const {booleanAttribute, stringAttribute} = require('../shared/attributes.js');
 
 const {Event} = require('../interface/event.js');
 const {Element} = require('../interface/element.js');
-const {Classes, customElements} = require('../interface/custom-element-registry.js');
+const {Classes, customElements, upgradingElements } = require('../interface/custom-element-registry.js');
 
 const Level0 = new WeakMap;
 const level0 = {
@@ -32,8 +32,15 @@ class HTMLElement extends Element {
 
   constructor(ownerDocument = null, localName = '') {
     super(ownerDocument, localName);
+    const {constructor: Class, [END]: end} = this;
+    if (upgradingElements.has(Class)) {
+      const {element, values } = upgradingElements.get(Class);
+      upgradingElements.delete(Class);
+      for (const [key, value] of values)
+        element[key] = value;
+      return element;
+    }
     if (!ownerDocument) {
-      const {constructor: Class, [END]: end} = this;
       if (!Classes.has(Class))
         throw new Error('unable to initialize this Custom Element');
       const {ownerDocument, localName, options} = Classes.get(Class);
--- a/esm/html/element.js
+++ b/esm/html/element.js
@@ -3,7 +3,7 @@ import {booleanAttribute, stringAttribute} from '../shared/attributes.js';
 
 import {Event} from '../interface/event.js';
 import {Element} from '../interface/element.js';
-import {Classes, customElements} from '../interface/custom-element-registry.js';
+import {Classes, customElements, upgradingElements } from '../interface/custom-element-registry.js';
 
 const Level0 = new WeakMap;
 const level0 = {
@@ -31,8 +31,15 @@ export class HTMLElement extends Element {
 
   constructor(ownerDocument = null, localName = '') {
     super(ownerDocument, localName);
+    const {constructor: Class, [END]: end} = this;
+    if (upgradingElements.has(Class)) {
+      const {element, values } = upgradingElements.get(Class);
+      upgradingElements.delete(Class);
+      for (const [key, value] of values)
+        element[key] = value;
+      return element;
+    }
     if (!ownerDocument) {
-      const {constructor: Class, [END]: end} = this;
       if (!Classes.has(Class))
         throw new Error('unable to initialize this Custom Element');
       const {ownerDocument, localName, options} = Classes.get(Class);
--- a/esm/interface/custom-element-registry.js
+++ b/esm/interface/custom-element-registry.js
@@ -8,6 +8,8 @@ export const Classes = new WeakMap;
 
 export const customElements = new WeakMap;
 
+export const upgradingElements = new WeakMap;
+
 export const attributeChangedCallback = (element, attributeName, oldValue, newValue) => {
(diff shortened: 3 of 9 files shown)


WebReflection:
All my libraries work with this project without issues ... and I am not sure why lit-element internals should compromise performance and logic for everything else in the SSR world.

However, I'd be happy to have a look as soon as I see a **test case** for this issue, or a better explanation of what is it that lit-element needs and why ... I am usually more keen to provide yet another dedicated export here, because I don't want to specialize this tool for libraries that needs SSR workarounds or it'd be a maintenance hell.

WebReflection:
> Lit itself is just a thin wrapper around custom element callbacks without any hacks to my knowledge.

trust me, you don't need to explain to me what lit is ... I've published lit patterns before lit existed ;-)

esm/html/element.js
```diff
@@ -31,8 +31,15 @@ export class HTMLElement extends Element {
 
   constructor(ownerDocument = null, localName = '') {
     super(ownerDocument, localName);
+    const {constructor: Class, [END]: end} = this;
+    if (upgradingElements.has(Class)) {
+      const {element, values } = upgradingElements.get(Class);
+      upgradingElements.delete(Class);
```
WebReflection, line 37: could you please reason this?

WebReflection: let me try again ... could you please reason this?

```js
upgradingElements.delete(Class);
```

why would you drop the class? what about further "same class" elements?

tmaroschik: I'm using the `upgradingElements` `WeakMap` as context object between the `HTMLElement` and the `CustomElementRegistry`, as its imho the best way to pass objects around between modules and not preventing garbage collection. In contrast in the webcomponents polyfill they used a module scoped variable to track the element that is currently beeing upgraded. 

`WeakMaps` require an object as key, so it felt natural to me to use the `Class` of the custom element, because the `Class` is also known in the `HTMLElement` constructor. As the `value` of the map entry I needed to use the currently upgrading `element` that needs to be returned inside the constructor. 

The value of the `upgradingElements` entry with the class key is always a single particular instance of an `HTMLElement` in the DOM. If I wouldn't delete the entry from the map, an instantiation like `new CustomElement()` would reference the last upgrading element instance related to that custom element class, which leads to very nasty bugs.

The assumptions here are that upgrading a single element happens synchronously and every custom element is a child of `HTMLElement`.

esm/interface/custom-element-registry.js
```diff
@@ -146,11 +148,11 @@ export class CustomElementRegistry {
         for (const [key] of values)
           delete element[key];
 
-        setPrototypeOf(element, new Class(this.ownerDocument, ce));
+        setPrototypeOf(element, Class.prototype);
         customElements.set(element, {connected: isConnected});
 
-        for (const [key, value] of values)
-          element[key] = value;
```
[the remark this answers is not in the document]
WebReflection: OK, I think I am getting what's going on ... so, if you move this part into the HTMLElement constructor it has the side effect that `onconnectedCallback` will be callbed **before** accessors / properties are set, which is a violation of the standard or not how *upgrade* works.

I'll try to think about the issue ... if I understood what it is: basically the `new Class` as prototype interfere with lit because it can't trap the element right away, only an instance that is *not* actually the current element ... right? Sorry if I am being slow

esm/interface/custom-element-registry.js
```diff
@@ -146,11 +148,11 @@ export class CustomElementRegistry {
...
+        upgradingElements.set(Class, { element, values });
+        new Class(this.ownerDocument, ce)
```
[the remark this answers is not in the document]
WebReflection: Never mind, I've got what you are trying to do.

cjs/html/element.js
```diff
@@ -32,8 +32,15 @@ class HTMLElement extends Element {
 
   constructor(ownerDocument = null, localName = '') {
     super(ownerDocument, localName);
+    const {constructor: Class, [END]: end} = this;
+    if (upgradingElements.has(Class)) {
```
WebReflection, line 36: this has a performance impact for everything ... from zero `.has` checks, we now pass through a WeakMap for every created element. I don't like it. This whole project degrades performance on demand: you use MutationObserver? One level perf degradation (inevitable) ... you use the Custom Elements Registry? Another level of perf degradation.

Accordingly, I would probably consider something like this instead:

```js

  constructor(ownerDocument = null, localName = '') {
    super(ownerDocument, localName);
    const {constructor: Class, [END]: end} = this;
    if (!ownerDocument) {
      if (!Classes.has(Class))
        throw new Error('unable to initialize this Custom Element');
      const {ownerDocument, localName, options} = Classes.get(Class);
      this.ownerDocument = end.ownerDocument = ownerDocument;
      this.localName = localName;
      customElements.set(this, {connected: false});
      if (options.is)
        this.setAttribute('is', options.is);
    }
    else {
      const {[CUSTOM_ELEMENTS]: {active}} = ownerDocument;
      if (active && upgradingElements.has(Class)) {
        // ... that thing
      }
    }
  }
```

However, I am not even sure if that's faster or slower, but that's how it's done for MutationObserver too so it's surely more consistent.

WebReflection:
@tmaroschik I think [I've done it](https://github.com/WebReflection/linkedom/pull/110) and please have a look. It's slightly simpler, imho, and it has less performance implication than your proposal. I could've asked you to write that but [I have limited time these days](https://twitter.com/WebReflection/status/1490618596271267840) so I hope you don't mind. I will merge that regardless because indeed the issue is not a *lit* only thing, it's an issue in general, and as ugly as the workaround is, it's necessary there.

Thanks for your patience 👋

WebReflection:
P.S. it's up and running 🥳

Issue #106, opened 2022-02-09
DOMParser.parseFromString(...) does not insert HTML into <html><body> document boilerplate


MadLittleMods:
`DOMParser.parseFromString(...)` does not insert HTML into the `<html><body>...</body></html>` document boilerplate which makes grabbing nodes inside inconsistent from the browser.

Tested in Chrome and Firefox:

```js
new DOMParser().parseFromString(`<div>asdf</div>`, "text/html").documentElement.outerHTML
// '<html><head></head><body><div>asdf</div></body></html>'

new DOMParser().parseFromString(`<div>asdf</div>`, "text/html").body.outerHTML
// '<body><div>asdf</div></body>'
```

`linkeddom`:

```js
import {DOMParser} from 'linkedom';

new DOMParser().parseFromString(`<div>asdf</div>`, "text/html").documentElement.outerHTML
// <div>asdf</div>

new DOMParser().parseFromString(`<div>asdf</div>`, "text/html").body.outerHTML
// <body></body>
```

---

For reference: I am using `linkedom` to server-side render (SSR) https://github.com/vector-im/hydrogen-web which uses the `DOMParser` for some of its templating, see [`src/platform/web/parsehtml.js#L67`](https://github.com/vector-im/hydrogen-web/blob/b8687343784070c53ca383f40ba1e10485121614/src/platform/web/parsehtml.js#L67).

WebReflection:
I don't really care about sloppy cases, but this one is kinda surprising 

```js
new DOMParser().parseFromString(`<html><body><div>asdf</div></body></html>`, "text/html").body.outerHTML
```

so eventually I'd fix this one only ... as all others are OK to me: you parse without outer nodes, you get what you passed. Pass outer elements instead and avoid surprises? One of the cases where explicit is better than implicit, and the "*but the standard does ...*" is not usually a strong argument in here :wink:

MadLittleMods:
Sounds good, just looking for a consistent way between both environments 👍

WebReflection:
The `body` case with missing `head` has been fixed. The rest is people writing sloppy code, I don't have much interest in sanitizing lazy developers ... I hope this is enough

MadLittleMods:
Related to https://github.com/WebReflection/linkedom/issues/160

WebReflection:
@MadLittleMods that has been fixed there https://github.com/WebReflection/linkedom/issues/160#issuecomment-1249088772

PR #108, opened 2022-02-09
Fix document.body not including content from DOMParser parsed <body>

MadLittleMods:
Fix `document.body` not including content from `DOMParser` parsed `<body>`

Fix https://github.com/WebReflection/linkedom/issues/106

Before:
```js
new DOMParser().parseFromString(`<html><body><div>asdf</div></body></html>`, "text/html").body.outerHTML
// '<body></body>'
```

After:
```js
new DOMParser().parseFromString(`<html><body><div>asdf</div></body></html>`, "text/html").body.outerHTML
// '<body><div>asdf</div></body>'
```

### DIFF
--- a/esm/shared/parse-from-string.js
+++ b/esm/shared/parse-from-string.js
@@ -57,12 +57,7 @@ export const parseFromString = (document, isHTML, markupLanguage) => {
     onopentag(name, attributes) {
       let create = true;
       if (isHTML) {
-        if (name === 'body') {
-          // Accessing the document.body property will create the <body>
-          node = document.body;
-          create = false;
-        }
-        else if (ownerSVGElement) {
+        if (ownerSVGElement) {
           node = append(node, document.createElementNS(SVG_NAMESPACE, name), active);
           node.ownerSVGElement = ownerSVGElement;
           create = false;
--- a/esm/shared/parse-from-string.js
+++ b/esm/shared/parse-from-string.js
@@ -57,7 +57,12 @@ export const parseFromString = (document, isHTML, markupLanguage) => {
     onopentag(name, attributes) {
       let create = true;
       if (isHTML) {
-        if (ownerSVGElement) {
+        if (name === 'body') {
+          // Accessing the document.body property will create the <body>
+          node = document.body;
+          create = false;
+        }
+        else if (ownerSVGElement) {
           node = append(node, document.createElementNS(SVG_NAMESPACE, name), active);
           node.ownerSVGElement = ownerSVGElement;
           create = false;
--- a/test/html/document.js
+++ b/test/html/document.js
@@ -48,6 +48,11 @@ assert(document.all[1], document.querySelector('head'));
 assert(document.all[2], document.querySelector('title'));
 assert(document.all[3], document.querySelector('body'));
 
+document = (new DOMParser).parseFromString('<!DOCTYPE html><html><body><div>foo</div></body></html>', 'text/html');
+assert(document.body.tagName, 'BODY');
+assert(document.body, document.querySelector('body'));
+assert(document.body.textContent, 'foo');
+
 // global listener
 let triggered = false;
 window.addEventListener('test', function once() {
--- a/test/html/document.js
+++ b/test/html/document.js
@@ -38,6 +38,8 @@ document.title = '"hello"';
 assert(document.title, '"hello"', 'title not escaped');
 assert(document.toString(), '<!DOCTYPE html><html><head><title>"hello"</title></head></html>');
 assert(document.body.tagName, 'BODY');
+assert(document.head, document.querySelector('head'));
+assert(document.head.textContent, '"hello"');
 
 document.title = '&';
 assert(document.toString(), '<!DOCTYPE html><html><head><title>&</title></head><body></body></html>');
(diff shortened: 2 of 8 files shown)


test/html/document.js
```diff
@@ -48,6 +48,11 @@ assert(document.all[1], document.querySelector('head'));
 assert(document.all[2], document.querySelector('title'));
 assert(document.all[3], document.querySelector('body'));
 
+document = (new DOMParser).parseFromString('<!DOCTYPE html><html><body><div>foo</div></body></html>', 'text/html');
+assert(document.body.tagName, 'BODY');
```
MadLittleMods, line 52: Making sure `document.body` isn't `null`

test/html/document.js
```diff
@@ -48,6 +48,11 @@ assert(document.all[1], document.querySelector('head'));
...
+assert(document.body, document.querySelector('body'));
```
MadLittleMods, line 53: Making sure `document.body` resolves to the same `Node` as `document.querySelector('body')`

test/html/document.js
```diff
@@ -48,6 +48,11 @@ assert(document.all[1], document.querySelector('head'));
...
+assert(document.body.textContent, 'foo');
```
MadLittleMods, line 54: Making sure the `document.body` includes the content we parsed.

esm/shared/parse-from-string.js
```diff
@@ -57,7 +57,12 @@ export const parseFromString = (document, isHTML, markupLanguage) => {
     onopentag(name, attributes) {
       let create = true;
       if (isHTML) {
-        if (ownerSVGElement) {
+        if (name === 'body') {
+          // Accessing the document.body property will create the <body>
+          node = document.body;
+          create = false;
+        }
```
MadLittleMods, line 64: The HTML spec says "The body element of a document is the first of [the `html` element](https://html.spec.whatwg.org/multipage/dom.html#the-html-element-2)'s children that is either a [`body`](https://html.spec.whatwg.org/multipage/sections.html#the-body-element) element or a [`frameset`](https://html.spec.whatwg.org/multipage/obsolete.html#frameset) element, or null if there is no such element." -- https://html.spec.whatwg.org/multipage/dom.html#dom-document-body-dev

But in terms of multiple `<body>` and how it behaves, the browser seems to resolve to the same `Node`.

```js
new DOMParser().parseFromString('<html><body>foo</body><body>bar</body></html>', "text/html").documentElement.outerHTML
// '<html><head></head><body>foobar</body></html>'
```

For reference, `linkedom` and the browser both behave the same here ✅

WebReflection: still it's no goal of this project to care about these shenanigans ... but good to know, thanks

WebReflection:
`Fix #106 - ... ` is the right commit message, so it relates to the issue, thanks!

MadLittleMods:
@WebReflection Updated ✅

esm/shared/parse-from-string.js
MadLittleMods: In this case, it's the simplest way to support and turns out to behave identically 👌

WebReflection:
Build failing. This MR is ignoring SVG and XML constraints, plus it's not considering that the current logic works already for `document.body` so it's kinda patching it in the wrong place.

The `body` getter is likely *the* place to look for a change, not the `parse-from-string` one, as that file has zero responsibility on how that `document` or its getters should react .. and bear in mind `document.head` is another one to eventually fix in the original issue.

worker.js
```diff
@@ -16692,7 +16692,7 @@ class HTMLDocument extends Document$1 {
   get head() {
     const {documentElement} = this;
     let {firstElementChild} = documentElement;
-    if (!firstElementChild) {
+    if (!firstElementChild || firstElementChild.tagName !== 'HEAD') {
```
MadLittleMods, line 16695: > [...] plus it's not considering that the current logic works already for document.body so it's kinda patching it in the wrong place.
> 
> The body getter is likely the place to look for a change, not the parse-from-string one, as that file has zero responsibility on how that document or its getters should react .. and bear in mind document.head is another one to eventually fix in the original issue.

Is this more what you mean?

---

Previously, the code would always assume the that first child was a `<head>` element but in the case of `<!DOCTYPE html><html><body><div>foo</div></body></html>`, it's a `<body>`. This means that when the `body` getter calls the `head` getter, it returned the existing `<body>` and created another `<body>` for itself after the supposed `<head>`.

The new code makes sure that the first child is a `<head>` element otherwise creates one.

This current assumptions about node order here has the downside that if someone defines stuff out of order, it will have 2 `<head>` when you try to access `document.head`

```js
// notice <head> after <body>
const document = new DOMParser().parseFromString(`<!DOCTYPE html><html><body></body><head><title>"hello"</title></head></html>`, "text/html").head.outerHTML
// '<head></head>'
```

We could fix this up better in `parse-from-string` if we enforce `<head>` as the first child and `<body>` after when documents are initially parsed and created. But I'd go back to the `document.body` creation pattern like I had previously to keep things simple.

WebReflection:
Fixed with latest.

MadLittleMods:
Thanks for incorporating the changes @WebReflection 😀

Looks pretty similar: https://github.com/WebReflection/linkedom/commit/535c5bc9508e15cfb858223d4fc9b7710c1bf54d

PR #111, opened 2022-02-10
fix: custom elements cannot be upgraded without ownerDocument arg

tmaroschik:
The current implementation fails to upgrade custom elements when no ownerDocument argument is passed, which is the case for most custom elements.

### DIFF
--- a/test/interface/custom-element-registry.js
+++ b/test/interface/custom-element-registry.js
@@ -2,7 +2,13 @@ const assert = require('../assert.js').for('CustomElementRegistry');
 
 const {parseHTML} = global[Symbol.for('linkedom')];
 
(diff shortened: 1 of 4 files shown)


test/interface/custom-element-registry.js
```diff
@@ -190,3 +196,12 @@ customElements.define('custom-element', class extends CE {
 })
 
 assert(document.querySelector('custom-element').isCurrentThisSameAsConstructorThis(), true, 'constructor this same as method this')
+
+const ceWithoutConstructorArgs = document.querySelector('custom-element-without-constructor-args');
+customElements.define('custom-element-without-constructor-args', class extends HTMLElement {
+  constructor() {
```
WebReflection, line 202: this constructor is pointless

Issue #109, opened 2022-02-09
Cannot read properties of undefined (reading 'clickfalse')


markwylde:
I'm having a problem with an app I'm building and have isolated it down to the following code snippet.

Basically, the steps to recreate are:
1. Create a new project
2. Init npm
3. Install `linkedom` and `preact`
4. Create an `index.js` file with the code below
5. Run `node index.js`
6. Receive the error `Cannot read properties of undefined (reading 'clickfalse')`

I can confirm the code works fine in Chrome. i.e. running `.click()` on the element.

## What is the code?
```javascript
import { parseHTML } from 'linkedom';
import { h, render } from 'preact';

function handleClick () {
  console.log('i gotted clicked')
}
const app = h('button', { onClick: handleClick }, 'Hello World!');

const {
  window, document
} = parseHTML(`
  <!doctype html>
  <html lang="en">
    <head>
      <title>Hello SSR</title>
    </head>
    <body>
    </body>
  </html>
`);

globalThis.document = document;
globalThis.window = window;

render(app, document.body);

const button = document.body.querySelector('button')
console.log(button.click())


```

## What is the error?
```
TypeError: Cannot read properties of undefined (reading 'clickfalse')
    at GlobalEvent.I (file:///home/mark/Synced/Projects/preact/node_modules/preact/dist/preact.mjs:1:4267)
    at GlobalEvent.dispatch (file:///home/mark/Synced/Projects/preact/node_modules/linkedom/esm/interface/event-target.js:7:14)
    at GlobalEvent.invokeListeners (file:///home/mark/Synced/Projects/preact/node_modules/linkedom/esm/interface/event-target.js:28:20)
    at Array.some (<anonymous>)
    at HTMLButtonElement.dispatchEvent (file:///home/mark/Synced/Projects/preact/node_modules/linkedom/esm/interface/event-target.js:80:17)
    at HTMLButtonElement.click (file:///home/mark/Synced/Projects/preact/node_modules/linkedom/esm/html/element.js:58:18)
    at file:///home/mark/Synced/Projects/preact/index.js:28:20
    at ModuleJob.run (node:internal/modules/esm/module_job:185:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:281:24)
```

WebReflection:
[...]

PR #112, opened 2022-02-10
Fix target on dispatch

markwylde:
Fixes https://github.com/WebReflection/linkedom/issues/109


WebReflection:
This has been fixed in a slightly different way. Thanks.

PR #114, opened 2022-02-12
Fix #113 - Add support for `CSSStyleDeclaration` property get, set, and remove

### LINKED ISSUE #113: Add support for `CSSStyleDeclaration.setProperty()` and `CSSStyleDeclaration.removeProperty()`
Add support for [`CSSStyleDeclaration.setProperty()`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty), [`CSSStyleDeclaration.removeProperty()`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/removeProperty), and [`CSSStyleDeclaration.getPropertyValue()`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/getPropertyValue).

These are essentially aliases for the [getters and setters already present on `esm/interface/css-style-declaration.js#L31-L61`](https://github.com/WebReflection/linkedom/blob/de941b7afd3e36765de12ef40ea1b36acaf0f0fd/esm/interface/css-style-declaration.js#L31-L61).

 - `HTMLElement.style.setProperty('background-color', 'red');` -> `HTMLElement.style.backgroundColor = 'red';` (existing `setter`)
 - `HTMLElement.style.removeProperty('background-color')` -> `HTMLElement.style.backgroundColor = null;` (existing `setter`)
 - `HTMLElement.style.getPropertyValue('background-color')` -> `HTMLElement.style.backgroundColor` (existing `getter`)

Support for these functions was also mentioned in https://github.com/WebReflection/linkedom/issues/92#issuecomment-919038796 where there is some consideration from @WebReflection about whether `linkedom` should support it.

---

For reference: I am using `linkedom` to server-side render (SSR) https://github.com/vector-im/hydrogen-web which uses `setProperty`, `removeProperty`, and `getPropertyValue` in a few places. SSR project is at: https://github.com/matrix-org/matrix-public-archive and `linkedom` usage is in [`server/render-hydrogen-to-string.js`](https://github.com/matrix-org/matrix-public-archive/blob/82ba92b8061d56d80b7b6cc3b3a61d25d28c1b93/server/render-hydrogen-to-string.js).

## Relevant code

 - [`esm/interface/css-style-declaration.js`](https://github.com/WebReflection/linkedom/blob/de941b7afd3e36765de12ef40ea1b36acaf0f0fd/esm/interface/css-style-declaration.js#L64-L112)

MadLittleMods:
Add support for `CSSStyleDeclaration` property get, set, and remove:

 - [`CSSStyleDeclaration.getPropertyValue()`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/getPropertyValue)
 - [`CSSStyleDeclaration.setProperty()`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty)
 -  [`CSSStyleDeclaration.removeProperty()`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/removeProperty)
 
Fix https://github.com/WebReflection/linkedom/issues/113

### DIFF
--- a/cjs/interface/css-style-declaration.js
+++ b/cjs/interface/css-style-declaration.js
@@ -82,6 +82,21 @@ class CSSStyleDeclaration extends Map {
     refs.get(this).setAttribute('style', value);
   }
 
+  getPropertyValue(name) {
+    const self = this[PRIVATE];
+    return handler.get(self, name);
+  }
+
+  setProperty(name, value) {
+    const self = this[PRIVATE];
+    handler.set(self, name, value);
+  }
+
+  removeProperty(name) {
+    const self = this[PRIVATE];
+    handler.set(self, name, null);
+  }
+
   [Symbol.iterator]() {
     const keys = getKeys(this[PRIVATE]);
     const {length} = keys;
--- a/esm/interface/css-style-declaration.js
+++ b/esm/interface/css-style-declaration.js
@@ -81,6 +81,21 @@ export class CSSStyleDeclaration extends Map {
     refs.get(this).setAttribute('style', value);
   }
 
+  getPropertyValue(name) {
+    const self = this[PRIVATE];
+    return handler.get(self, name);
+  }
+
+  setProperty(name, value) {
+    const self = this[PRIVATE];
+    handler.set(self, name, value);
+  }
+
+  removeProperty(name) {
+    const self = this[PRIVATE];
+    handler.set(self, name, null);
+  }
+
   [Symbol.iterator]() {
     const keys = getKeys(this[PRIVATE]);
     const {length} = keys;
--- a/test/interface/css-style-declaration.js
+++ b/test/interface/css-style-declaration.js
@@ -28,3 +28,9 @@ assert(node.toString(), '<div></div>', 'setter as null');
 node.id = '';
 node.className = '';
 assert(node.toString(), '<div></div>', 'setter as null');
+
+node.style.setProperty('background-color', 'purple');
+assert(node.toString(), '<div style="background-color:purple"></div>', 'setProperty');
+assert(node.style.getPropertyValue('background-color'), 'purple', 'getPropertyValue');
+node.style.removeProperty('background-color')
+assert(node.toString(), '<div></div>', 'removeProperty');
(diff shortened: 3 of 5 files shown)


cjs/interface/css-style-declaration.js
```diff
@@ -82,6 +82,21 @@ class CSSStyleDeclaration extends Map {
     refs.get(this).setAttribute('style', value);
   }
 
+  getPropertyValue(name) {
+    const self = this[PRIVATE];
```
MadLittleMods, line 86: For my own understanding/curiosity, what is the difference between `this` and `this[PRIVATE]` (which is a getter that returns `this`)?

It seems like the `cssText` getters and setters are the only ones that use `this`

WebReflection: `this` points at the proxy, not at the original `Map` reference ... the `constructor` returns a Proxy ;-)

esm/interface/css-style-declaration.js
```diff
@@ -81,6 +81,21 @@ export class CSSStyleDeclaration extends Map {
     refs.get(this).setAttribute('style', value);
   }
 
+  getPropertyValue(name) {
+    const self = this[PRIVATE];
+    return handler.get(self, name);
```
WebReflection, line 86: The handler is for proxy thing ... once you have the original *Map* that is holding all values, you gotta do like:

```js
return self.get(uhyphen(name));
```

got it?

esm/interface/css-style-declaration.js
```diff
@@ -81,6 +81,21 @@ export class CSSStyleDeclaration extends Map {
...
+  }
+
+  setProperty(name, value) {
+    const self = this[PRIVATE];
+    handler.set(self, name, value);
```
WebReflection, line 91: this is fine only because there's quite some logic in the handler ... when it comes to the getter there's nothing else than a *map.get*

esm/interface/css-style-declaration.js
WebReflection: Actually, never mind ... `updateKeys` needs to happen, so this was unnecessary optimization. All good!

cjs/interface/css-style-declaration.js
MadLittleMods: Ahhh, the proxy passes the `this[PRIVATE]` getter back to the original object and accesses that `this`. Thanks for the explanation 🙂

MadLittleMods:
Thanks for the review, merge, and release @WebReflection 🎉

PR #127, opened 2022-04-09
support shadow root nested custom elements

dchester:
Thanks for this wonderful library!  This PR makes sure to fire `connectedCallback` and `disconnectedCallback` even for custom elements contained within shadow roots of other custom elements.  

When recursively firing `connectedCallback` and `disconnectedCallback` we need to traverse down through any shadow roots we find along the way.

And then when determining `isConnected` for a node within a shadow root, we need to traverse up through shadow root host elements to find our way to the owner document.  In order to facilitate this, we implement the [`host`](https://dom.spec.whatwg.org/#dom-shadowroot-host) property for shadow roots.

### DIFF
--- a/cjs/interface/custom-element-registry.js
+++ b/cjs/interface/custom-element-registry.js
@@ -38,7 +38,7 @@ const triggerConnected = createTrigger('connectedCallback', true);
 const connectedCallback = element => {
   if (reactive) {
     triggerConnected(element);
-    let {[NEXT]: next, [END]: end} = element;
+    let {[NEXT]: next, [END]: end} = element.shadowRoot || element;
     while (next !== end) {
       if (next.nodeType === ELEMENT_NODE)
         triggerConnected(next);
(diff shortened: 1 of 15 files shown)


WebReflection:
@dchester nice and clean PR, good stuff! My only concern is around the closed mode ... I think a WeakMap with all element -> shadowRoot nodes, instead of trusting that SR will always be open and reachable, might be a better way forward ... what do you think?

dchester:
Yes, good point.  I'll try that out.

WebReflection:
Btw, this project is more for SSR than for testing headless pages, which is why I never cared much about shadowRoot, because it cannot be serialized as SSR (yet? who knows when) ... so please keep in mind as soon as shadow roots will be serializable as SSR the code around it might change.

dchester:
I pushed up a commit that does better now with closed shadow roots.

It looks like there's already a `shadowRoots` WeakMap in place in `element.js` to help with this, but I had to extract it to its own module to avoid circular dependencies.  Not sure if there's a better way, but this at least does the trick.  What do you think?

dchester:
Yes, makes sense.  Thanks much!

PR #129, opened 2022-04-14
Handle `style` values containing `:`

delucis:
Certain values in `style` can contain colons, for example: `background-image: url('https://example.com/image.png');`.

The current logic discards the colon and everything that follows in the URL, resulting in a value of `url('https` for `style['background-image']`.

This commit fixes that by reconstituting values after the first `:` in a rule.


esm/interface/css-style-declaration.js
```diff
@@ -13,9 +13,9 @@ const updateKeys = style => {
     if (attr) {
       style.set(PRIVATE, attr);
       for (const rule of attr[VALUE].split(/\s*;\s*/)) {
-        const pair = rule.split(/\s*:\s*/);
-        if (1 < pair.length) {
-          let [key, value] = pair;
+        const [key, ...rest] = rule.split(':');
+        const value = rest.join(':');
```
WebReflection, line 17: do you mind moving this *after* the `if` check? 'cause it's a completely pointless operation if `rest.length` is `0`, right?

WebReflection: on top of that, value should be `rest.join(':').trim()` within the `if` 😉 ... also, if you use `/\s*:\s*/` in the split the `trim` operation will be faster because the split already took care of dropping spaces, and I don't think there are cases with meant spaces around colons, including URLs that are badly formatted if that's the case

delucis: Made that optimisation.

There are cases where spaces around a colon could be relevant like with `content`. Do you mind keeping it as is? My guess is that the RegExp creation might be just as expensive as trimming additional space on one end the string.

WebReflection: The static RegExp can be easily optimized by engines, but fair enough about the content case, although the trim is still needed

delucis: Thanks. Then I think this is ready ☺️

Do you need a test case for this?

delucis: P.S. really nice codebase — used this for the first time this afternoon, hit the error and could find and fix it super easily.

WebReflection: Ideally I’d like a test case as the code is covered but your use case is not double checked

delucis:
Added a test case for this. I'm on my phone so I couldn't run it, but hopefully I understood the code correctly 😬

esm/interface/css-style-declaration.js
```diff
@@ -13,11 +13,10 @@ const updateKeys = style => {
     if (attr) {
       style.set(PRIVATE, attr);
       for (const rule of attr[VALUE].split(/\s*;\s*/)) {
-        const pair = rule.split(/\s*:\s*/);
-        if (1 < pair.length) {
-          let [key, value] = pair;
+        const [key, ...rest] = rule.split(':');
```
WebReflection, line 16: in two lines we have `key = key.trim()` which is meant to be there because:

```js
document.documentElement.style.cssText = 'background-color : white';
```

I hence believe this `const [...]` should be a `let [...]` as it was before, or the tests fail and nothing lands.

delucis: Of course! Sorry about that. Should be fixed.

esm/interface/css-style-declaration.js
```diff
@@ -13,11 +13,10 @@ const updateKeys = style => {
...
       style.set(PRIVATE, attr);
       for (const rule of attr[VALUE].split(/\s*;\s*/)) {
-        const pair = rule.split(/\s*:\s*/);
-        if (1 < pair.length) {
-          let [key, value] = pair;
+        let [key, ...value] = rule.split(':');
+        if (value.length > 0) {
           key = key.trim();
-          value = value.trim();
+          value = value.join(':').trim();
```
WebReflection, line 19: not sure why I've asked to make `key` mutable and we ended up with `value` changing type at runtime, which is never desired perf-wise-speaking 🤔

please rollback this last change and just make `const` a `let` or put back the `rest` reference without changing references type at runtime, thanks!

delucis: My bad, sorry — not too familiar with these performance considerations. Thanks for the very thorough reviews 🙏

WebReflection:
@delucis thanks! now we have another issue ... the check fails, and it's probably because we're incapable of splitting keys into multiple styles ... so this is some sort of a bummer, and it needs a completely different fix from this one, imho.

All strings should rather be dropped and then re-inserted, so that the semicolon does its duty, without interfering with strings.

I think I can take care of this, but I won't be too fast these days (family here, new born, etc).

If you find a way to fix this though, I'll be happy to review, otherwise feel free to file an issue and I'll work on that when I can, thanks.

WebReflection:
**actually** ... never mind, I confused `:` with `;` ... so, definitively we need to understand why the test fails now, as it shouldn't.

It might be not a regression, but somethin ain't right ... the split on `;` does not produce the expected result ... it's possible

delucis:
Fixed! Sorry working on my phone yesterday was a bad idea — needed to update several other test cases. Got tests passing locally now so 🤞 should be OK.

WebReflection:
Up and running: thank you 🥳

delucis:
Fantastic! Thanks for your patience with me on this one & have a great weekend.

WebReflection:
@delucis nope, thank **you** for fixing something I didn't care about too much: it was a good call 🍻

Issue #134, opened 2022-05-02
setAttribute/removeAttribute not clearing cache


aghull:
When using the cached version, attribute changes from adding or removing attributes do not reset the cache and allow stale cache results to show up. E.g. given a `node` with children:
```
assert(node.querySelector('[c]'), null);
node.childNodes[0].setAttribute('c', '3');
assert(node.querySelector('[c]'), node.childNodes[0]); // FAIL
```
The above fails with the cached empty result

```
node.childNodes[0].setAttribute('d', '4');
assert(node.querySelector('[d]'), node.childNodes[0]);
node.childNodes[0].removeAttribute('d');
assert(node.querySelector('[d]'), null); // FAIL
```
The above fails with the cached non-empty result

WebReflection:
This indeed looks like a bug. Any particular reason you’re using the cached version, if I might ask?

WebReflection:
Never mind, I've patched the issue for all missing methods too now.

aghull:
I'm using it for a project that does a lot of XML querying and manipulation and wanted a really fast QSA. The caching would help me a lot and if I don't use this, I'd probably roll my own.

PR #135, opened 2022-05-02
Reset cache on setAttribute/removeAttribute

aghull:
Addresses issue #134 

Haven't worked on this code before so hopefully this isn't too messy. I don't see any tests that really cover all the caching/uncaching cases so unsure what the best testing strategy is here.


WebReflection:
I've patched the issue with all other methods too (setAttributeNS, setAttributeNode, and friends). Thanks 👋

PR #141, opened 2022-06-05
fix: `:scope` selector

### LINKED ISSUE #75: querySelector with :scope failing unexpectedly
Re-using the example from the readme, the following minimal example failed unexpectedly:

```js
const {parseHTML} = require ('linkedom');

const {
 document
  } = parseHTML(`
    <!doctype html>
    <html lang="en">
      <head>
        <title>Hello SSR</title>
      </head>
      <body>
        <form>
          <input name="user">
          <button>
            Submit
          </button>
        </form>
      </body>
    </html>
  `);


const form = document.querySelector('form');
const button = form.querySelector(':scope > button');
console.log(button); // returns null while DOM would return the button
```

I realize that this might be a bug in `css-select` but I don't know it well enough to figure things out.

blake-regalia:
Very simple fix to add the element as the context for css-select calls which fixes all issues related to using `:scope` selector.
Closes #75

### DIFF
--- a/esm/mixin/parent-node.js
+++ b/esm/mixin/parent-node.js
@@ -179,7 +179,7 @@ export class ParentNode extends Node {
   }
 
   querySelector(selectors) {
-    const matches = prepareMatch(this, selectors);
+    const matches = prepareMatch(this, selectors, this);
     let {[NEXT]: next, [END]: end} = this;
     while (next !== end) {
       if (next.nodeType === ELEMENT_NODE && matches(next))
@@ -190,7 +190,7 @@ export class ParentNode extends Node {
   }
 
   querySelectorAll(selectors) {
-    const matches = prepareMatch(this, selectors);
+    const matches = prepareMatch(this, selectors, this);
     const elements = new NodeList;
     let {[NEXT]: next, [END]: end} = this;
     while (next !== end) {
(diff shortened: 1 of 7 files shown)


esm/mixin/parent-node.js
```diff
@@ -179,7 +179,7 @@ export class ParentNode extends Node {
   }
 
   querySelector(selectors) {
-    const matches = prepareMatch(this, selectors);
+    const matches = prepareMatch(this, selectors, this);
```
WebReflection, line 182: passing `this` twice because ... ???

WebReflection:
P.S. also please write a test for `:scope` otherwise who knows if this solved anything or not and, if it does, who knows if somebody decides it's a good idea to not pass `context` anymore in the future, thank you!

WebReflection:
@blake-regalia OK, this MR was breaking tests because when `context` is used both `matches` and `closest` ignore looking at the element itself, which is annoying ... [my change](https://github.com/WebReflection/linkedom/commit/1ec427dc66cb68e62ad4626fad8f9ba926487152) made `:scope` arbitrary and added a test to keep code-coverage 100%.

Thanks for the hint though, I hope latest patch is good enough, even if not perfect. P.S. combined selectors with `:scope` will switch to context mode and results might be different from the real Web/browser one ... watch out!

PR #142, opened 2022-06-08
fix: scope selector in union

blake-regalia:
Adds broader support for `:scope`, specifically in union selector queries.

Still not perfect as someone could potentially have a query like `div[data-attr="fake,:scope>query"]` but that's very obscure.

Something else learned during integration is that the current `:root` selector is not perfect as it fails union queries (see commented out test case added in commit).

### DIFF
--- a/esm/shared/utils.js
+++ b/esm/shared/utils.js
@@ -8,6 +8,8 @@ export const getEnd = node => node.nodeType === ELEMENT_NODE ? node[END] : node;
 
 export const ignoreCase = ({ownerDocument}) => ownerDocument[MIME].ignoreCase;
 
+export const usesScopePseudoClass = (selectors) => selectors.split(/,/g).some(selector => /^\s*:(scope)\b/.test(selector));
+
 export const knownAdjacent = (prev, next) => {
   prev[NEXT] = next;
   next[PREV] = prev;
--- a/cjs/shared/matches.js
+++ b/cjs/shared/matches.js
@@ -2,7 +2,7 @@
 const CSSselect = require('css-select');
 
 const {ELEMENT_NODE, TEXT_NODE} = require('./constants.js');
-const {ignoreCase} = require('./utils.js');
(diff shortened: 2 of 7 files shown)


esm/shared/utils.js
```diff
@@ -8,6 +8,8 @@ export const getEnd = node => node.nodeType === ELEMENT_NODE ? node[END] : node;
 
 export const ignoreCase = ({ownerDocument}) => ownerDocument[MIME].ignoreCase;
 
+export const usesScopePseudoClass = (selectors) => selectors.split(/,/g).some(selector => /^\s*:(scope)\b/.test(selector));
```
WebReflection, line 11: what is this doing differently from `.includes(':scope')` beside being slower and greedier on both heap and garbage collector?

why are you also capturing `scope` in there?

wouldn't this work the same if not better?

```js
// no need to split, no need to some(...)
/(?:^|,)\s*:scope\b/.test(selectors)
```

blake-regalia: String matching scope is a hack. And perf is totally negligible, man but do whatever you want.

blake-regalia: No worries man, I'm not offended and I don't care. Fixing scope/root in your lib is just too much work anyway, that's why I closed PR. All of these will fail or cause issues for example:
 - `div[data-attr="p,:scope"]`
 - `[any],:root`
 - `:not(:root) div`

WebReflection: you do realize the issue is not here, as the CSS selector engine is third party, right? just in case you'd like to help solving the problem in the right place: https://github.com/fb55/css-select

blake-regalia: The issue is here, it is with how you handle the selectors string and provide `context` to css-select. Feel free to look thru [other libs](https://www.npmjs.com/package/css-select) that leverage css-select and implement selectors correctly if you need inspiration, [cheerio-select](https://github.com/cheeriojs/cheerio-select) is one. Good luck!

PR #144, opened 2022-06-19
Added sciprt string accessor

Innei:


esm/html/script-element.js
```diff
@@ -1,18 +1,96 @@
...
+  get innerText() {
+    return super.innerText || ''
+  }
+  set innerText(value) {
+    super.textContent = value
+  }
+  get textContent() {
+    return super.textContent || ''
+  }
+  set textContent(value) {
```
WebReflection, line 32: these look all redundant ... if you have to use getters and setters from super just *don't* implement that?

esm/html/script-element.js
```diff
@@ -1,18 +1,96 @@
...
-const tagName = 'script';
+const tagName = 'script'
 
 /**
  * @implements globalThis.HTMLScriptElement
  */
 class HTMLScriptElement extends TextElement {
   constructor(ownerDocument, localName = tagName) {
-    super(ownerDocument, localName);
+    super(ownerDocument, localName)
```
WebReflection, line 13: it'd be weird to have a single file with no semicolons and I think the linter would complain too ... so please use semicolons **everywhere** thank you!

PR #154, opened 2022-08-17
Add `urlAttribute` helper that uses `baseURI` (used for `href`, `src` etc.)

### LINKED ISSUE #153: `baseURI` should be factored into HTMLAnchorElement.href
[`Node.baseURI`](https://developer.mozilla.org/en-US/docs/Web/API/Node/baseURI) should be used when getting the `HTMLAnchorElement.href` property. (It maybe needed for other properties as well, will add them here once I investigate.)

danburzo:
Fixes #153 by adding an `urlAttribute` helper to use in IDL properties that need to reflect HTML attributes representing an URL, such as `HTMLAnchorElement.href`, `HTMLImageElement.src`, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ document.querySelectorAll('form, input[name], button');
 
 ```js
 // any node can be serialized
-const array = document.toJSON();
+const array = toJSON(document);
 
 // somewhere else ...
 import {parseJSON} from 'linkedom';


esm/shared/attributes.js
```diff
@@ -89,6 +89,23 @@ export const stringAttribute = {
   }
 };
 
+export const urlAttribute = {
+  get(element, name) {
+    let attr = element.getAttribute(name);
```
WebReflection, line 94: `const` ?

esm/shared/attributes.js
```diff
@@ -89,6 +89,23 @@ export const stringAttribute = {
...
+    if (attr) {
+      try {
+        return new URL(attr, element.baseURI).href;
```
WebReflection, line 97: in which case this could fail, if I might ask? 🤔

WebReflection: I guess my wonder is why this wouldn't be enough:

```js
export const urlAttribute = {
  get(element, name) {
    const attr = element.getAttribute(name);
    return attr ? new URL(attr, element.baseURI).href : '';
  },
  set(element, name, value) {
    element.setAttribute(name, value);
  }
};
```

but for unknown crawling maybe your version is safer ... although is the result expected at that point?

esm/html/anchor-element.js
```diff
@@ -14,8 +14,8 @@ class HTMLAnchorElement extends HTMLElement {
   }
 
   /* c8 ignore start */ // copy paste from img.src, already covered
-  get href() { return encodeURI(stringAttribute.get(this, 'href')); }
```
danburzo, line 17: The getter/setter for `href` used to encode/decode the URI, and this change breaks some tests. Should this technique be adopted inside `urlAttribute`? 

Browser behavior is below:

```js
let a = document.createElement('a');
a.href = 'https://google.com/?q=asd&lol=<2>';
a.href
// "https://google.com/?q=asd&lol=%3C2%3E"
a.getAttribute('href')
// "https://google.com/?q=asd&lol=<2>"
```

Should we also store the raw value in the attribute, and encode it on the getter?

WebReflection: if tests break it means we'll break working code out there ... this change shouldn't require a major though, should it?

WebReflection: P.S. yes, we might want to `encodeURI` on the URL returned href, if URL returned href is not encoded already

WebReflection: ... which it does ...


```js
new URL('/?q=asd&lol=<2>', 'https://google.com/').href 
```

so we want this behavior, I would drop the encoding but I am not sure we should drop the decoding ... the encoding is provided by the URL on current attribute which is the raw value, the decoding in assignment looks like still necessary to preserve the raw value ... this should not break anything, just remove the encodeURI in the getter.

danburzo: I need to read a bit more to understand how to align with the HTML spec (if possible). The immediate fix for the breaking test is:

```js
return new URL(attr, element.baseURI ?? undefined).href;
```

(`baseURI` is `null` in the absence of an explicit `location` or `<base>` element, and that's an invalid base URL.)

But the problem is the getter returns either the encoded or unencoded value depending on the validity of `baseURI`, and I wonder if that's the case in browsers as well.

danburzo: Hmm, I guess browsers work similarly when `baseURI` can't be used:

```js
let a = document.createElement('a');
let b = document.createElement('base');
document.head.appendChild(b);
a.href = '/?q=asd&lol=<2>';

console.log(a.href, a.getAttribute('href'), a.baseURI); 
// /?q=asd&lol=<2> /?q=asd&lol=<2> about:newtab

b.href = 'https://github.com';
console.log(a.href, a.getAttribute('href'), a.baseURI); 
// https://github.com/?q=asd&lol=%3C2%3E /?q=asd&lol=<2> https://github.com/
```

I guess the remaining questions are: 

* What's the proper default value for `baseURI` in a headless environment? 
* If the default value remains `null`, should the default state cause the `href` attribute to be passed to the `URL()` constructor, or return the raw attribute value (as if baseURI was not usable)?

WebReflection: all I can say is that if this will be overly complicated, or full of slow paths too, I am not too keen to change everything because of this use case that makes little sense for headless browser ... it's also not the main use case for the project, to be honest.

danburzo: JSDOM, for example, [uses `about:blank`](https://github.com/jsdom/jsdom/blob/04f6c13f4a4d387c7fc979b8f62c6f68d8a0c639/test/api/options.js#L66-L73) as the default base URL, with results below:

```js
a.href = '/?q=asd&lol=<2>';
console.log(a.href);
// /?q=asd&lol=<2>
a.href = 'https://google.com/?q=asd&lol=<2>'
console.log(a.href);
// https://google.com/?q=asd&lol=%3C2%3E
```

That is:

* `new URL('/?q=asd&lol=<2>', 'about:blank').href; // => throws, return raw attribute
* `new URL('https://google.com/?q=asd&lol=<2>', 'about:blank').href; // => works, return encoded URL

WebReflection: `about:blank` sounds like a sensible default then for `baseURI` ... that also involves a `<base>` element knowledge somehow though, isn't it?

danburzo: The current implementation of `baseURI` is aware of `<base>` right now, but if speed is a concern this needs to be cached or otherwise optimized, so that might complicate things:

```js
  get baseURI() {
    const ownerDocument = this.nodeType === DOCUMENT_NODE ?
                            this : this.ownerDocument;
    if (ownerDocument) {
      const base = ownerDocument.querySelector('base');
      if (base)
        return base.getAttribute('href');

      const {location} = ownerDocument.defaultView;
      if (location)
        return location.href;
    }

    return null;
  }
```

Let me know if you think this feature is worth it in relation with your goals / use-cases for `linkedom`. This has been an interesting rabbit hole regardless!

danburzo: Yeah, I don't see a big problem with `return 'about:blank';` although it's technically a breaking change, as is I guess changing the behavior of URL-like element attributes. What I'm more concerned is how to get rid of the inefficient `querySelector` each time you request `href` on a link, and still update the value when you mutate the DOM.

WebReflection: The cached version of this module doesn't care about repeated qS/qSA operation if the DOM didn't mutate in the meanwhile, so if that's a concern, just use the cached version instead?

WebReflection: That is `linkedom/cached` import, instead of the regular one https://github.com/WebReflection/linkedom#are-childnodes-and-children-always-computed

esm/interface/node.js
```diff
@@ -66,7 +66,7 @@ export class Node extends EventTarget {
     const ownerDocument = this.nodeType === DOCUMENT_NODE ?
                             this : this.ownerDocument;
     if (ownerDocument) {
-      const base = ownerDocument.querySelector('base');
```
danburzo, line 69: `<base>` elements without a `href` attribute should be skipped here. The [algorithm](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#document-base-url) is a bit more complicated (allowing for relative `href`), but maybe that can be addressed more thoroughly in a future PR.

PR #162, opened 2022-09-20
fix attribute clone missing ownerElement

xeaone:
This fixes missing ownerElement when attribute is cloned. See included example.

```js
import { parseHTML } from './esm/index.js';

const w = parseHTML(`<html><body>

<select value="{{fruit=$value}}" each="{{[fruits,'f']}}"><option value="{{f}}">{{f}}</option></select>

</body></html>`);

// w.document.toString();
// console.log(w.document.body.innerHTML);

const notClone = w.document.body.firstElementChild.firstElementChild;
const notCloneAt = notClone.attributes[0];
console.log(notCloneAt.name, notCloneAt.value, notCloneAt.ownerElement.outerHTML);

const clone = w.document.body.firstElementChild.firstElementChild.cloneNode(true);
const cloneAt = clone.attributes[0];
console.log(cloneAt.name, cloneAt.value, cloneAt.ownerElement.outerHTML); // without a fix ownerElement is not assigned
```

### DIFF
--- a/esm/interface/element.js
+++ b/esm/interface/element.js
@@ -138,7 +138,7 @@ export class Element extends ParentNode {
 
 
   // <contentRelated>
-  get innerText() { 
+  get innerText() {
     const text = [];
     let {[NEXT]: next, [END]: end} = this;
     while (next !== end) {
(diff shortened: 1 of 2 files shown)


WebReflection:
al good, but I'd love to have a test too that shows your expectations here, as it's easy to have regressions otherwise, once this lands

xeaone:
Sounds good I will see about adding a test for it.

xeaone:
Okay I added a test but not sure if it is following test practice let me know.

WebReflection:
it's up and running, thank you 👋

PR #168, opened 2022-10-15
Improve TS typing

### LINKED ISSUE #167: Typescript type errors
Hi there,
I'm running into loads of type errors from linkedom when using the library in a typescript project.

These type errors appear to all be either missing properties or incorrect implementations of base class methods (due to parameter types or return types not matching).

The errors are too long to send but I will send a few:

```ts
node_modules/linkedom/types/dom/parser.d.ts:12:5 - error TS2416: Property 'parseFromString' in type 'DOMParser' is not assignable to the same property in base type 'DOMParser'.
  Type '<MIME extends "image/svg+xml" | "text/html" | "text/xml">(markupLanguage: string, mimeType: MIME) => { "text/html": HTMLDocument; "image/svg+xml": SVGDocument; "text/xml": XMLDocument; }[MIME]' is not assignable to type '(string: string, type: DOMParserSupportedType) => Document'.
    Type 'HTMLDocument | SVGDocument | XMLDocument' is not assignable to type 'Document'.
      Type 'HTMLDocument' is missing the following properties from type 'Document': URL, alinkColor, anchors, applets, and 181 more.

12     parseFromString<MIME extends keyof {
       ~~~~~~~~~~~~~~~

node_modules/linkedom/types/html/anchor-element.d.ts:4:14 - error TS2420: Class 'import("C:/Users/felis/home/projects/test-linkedom/node_modules/linkedom/types/html/anchor-element").HTMLAnchorElement' incorrectly implements interface 'HTMLAnchorElement'.
  Type 'HTMLAnchorElement' is missing the following properties from type 'HTMLAnchorElement': charset, coords, hreflang, name, and 132 more.

4 export class HTMLAnchorElement extends HTMLElement implements globalThis.HTMLAnchorElement {
               ~~~~~~~~~~~~~~~~~

...

node_modules/linkedom/types/html/document.d.ts:6:9 - error TS2416: Property 'all' in type 'HTMLDocument' is not assignable to the same property in base type 'HTMLDocument'.
  Property 'namedItem' is missing in type 'NodeList' but required in type 'HTMLAllCollection'.

6     get all(): NodeList;
          ~~~

  ../../../AppData/Roaming/npm/node_modules/typescript/lib/lib.dom.d.ts:5953:5
    5953     namedItem(name: string): HTMLCollection | Element | null;
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    'namedItem' is declared here.

node_modules/linkedom/types/html/document.d.ts:15:9 - error TS2416: Property 'title' in type 'HTMLDocument' is not assignable to the same property in base type 'HTMLDocument'.
  Type 'HTMLTitleElement' is not assignable to type 'string'.

15     set title(arg: HTMLTitleElement);
           ~~~~~

...
```

Workaround: `tsc --skipLibCheck`

It'd be great to get the types correct here so `--skipLibCheck` isn't needed. I would make a PR for this however it seems design decisions would need to be made to fix the types. I'm not sure if the types can be reconciled while still extending / implementing `globalThis.DOMParser`, `globalThis.HTMLElement`, `/globalThis\.HTML.*Element/`, etc.

jeremy-rifkin:
This is an attempt to fix #167

I do not believe it's correct to suggest the DOMParser implements `globalThis.DOMParser`, the `HTMLElement` implements `globalThis.HTMLElement`, etc.

I think these changes should be ok given the goal of the library is to be close but not too close to the actual dom API.
Perhaps one of the original authors can lend insight into the original motivation for these `implements` annotations.

Todo:
- [x] Fix remaining errors
- [ ] Add an automated test case to ensure the library can be imported without errors


jeremy-rifkin:
The last error to fix is
```ts
../types/mixin/parent-node.d.ts:28:5 - error TS1170: A computed property name in a type literal must refer to an expression whose type is a literal type or a 'unique symbol' type.

28     [typeof NEXT]: NodeStruct;
       ~~~~~~~~~~~~~

../types/mixin/parent-node.d.ts:29:5 - error TS1170: A computed property name in a type literal must refer to an expression whose type is a literal type or a 'unique symbol' type.

29     [typeof PREV]: NodeStruct;
       ~~~~~~~~~~~~~

../types/mixin/parent-node.d.ts:30:5 - error TS1170: A computed property name in a type literal must refer to an expression whose type is a literal type or a 'unique symbol' type.

30     [typeof START]: NodeStruct;
       ~~~~~~~~~~~~~~


Found 3 errors in the same file, starting at: ../types/mixin/parent-node.d.ts:28
```

jeremy-rifkin:
I think that is the right fix

WebReflection:
oh gosh ... all definition files [are automatically generated per each CI build step or locally](https://github.com/WebReflection/linkedom/blob/main/package.json#L19) before pushing to npm. Apologies I should've mentioned this as this PR is quite some effort if it was fixed manually per each generated file, but next release will re-create again all files and all changes will be lost.

The files you need to eventually change are in the esm folder, not in the types one, as that's all automation.

jeremy-rifkin:
Oh shoot thanks! This PR was fast - just a bunch of regex replaces - so nothing of value lost. I’ll make the correct changes when I’m back at my computer.

esm/interface/document.js
```diff
@@ -174,7 +171,7 @@ export class Document extends NonElementParentNode {
   createComment(textContent) { return new Comment(this, textContent); }
   createDocumentFragment() { return new DocumentFragment(this); }
   createDocumentType(name, publicId, systemId) { return new DocumentType(this, name, publicId, systemId); }
-  createElement(localName) { return new Element(this, localName); }
+  createElement(localName, options) { return new Element(this, localName); }
```
jeremy-rifkin, line 174: `HTMLDocument.createElement` takes a second parameter

esm/interface/node.js
```diff
@@ -97,8 +94,14 @@ export class Node extends EventTarget {
   get textContent() { return null; }
   set textContent(value) {}
   normalize() {}
-  cloneNode() { return null; }
-  contains() { return false; }
+  /**
+   * @param {boolean | undefined} deep
+   */
+  cloneNode(deep) { return null; }
```
jeremy-rifkin, line 100: `Document`, `Element`, and `NonElementParentNode` take a `deep` parameter

esm/interface/node.js
```diff
@@ -97,8 +94,14 @@ export class Node extends EventTarget {
...
+  /**
+   * @param {Node} node
+   */
+  contains(node) { return false; }
```
jeremy-rifkin, line 104: `ParentNode` is the only implementer and takes a `node` parameter

esm/mixin/parent-node.js
```diff
@@ -37,9 +37,9 @@ const insert = (parentNode, child, nodes) => {
 };
 
 /** @typedef { import('../interface/element.js').Element & {
-    [typeof NEXT]: NodeStruct,
-    [typeof PREV]: NodeStruct,
-    [typeof START]: NodeStruct,
+    [NEXT]: NodeStruct,
+    [PREV]: NodeStruct,
+    [START]: NodeStruct,
```
jeremy-rifkin, line 42: TS wants the literal value rather than the type of the literal

package.json
```diff
@@ -34,6 +34,7 @@
     "@rollup/plugin-commonjs": "^22.0.0",
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-node-resolve": "^13.3.0",
+    "@types/node": "^18.11.0",
```
jeremy-rifkin, line 37: `@types/node` is needed for types like `Set`

jeremy-rifkin:
I noticed `types/dom`, `types/html`, `types/interface`, `types/mixin`, `types/shared`, `types/svg`,`types/xml`, and `types/index.d.ts` were not being auto-generated and haven't seen modifications in a very long time and the structure is the same as `types/esm`. I deleted these stale files.

types/esm/interface/custom-event.d.ts
```diff
@@ -1,37 +1,2 @@
...
-        composedPath(): any[];
-        stopPropagation(): void;
-        stopImmediatePropagation(): void;
-    };
-    readonly BUBBLING_PHASE: number;
-    readonly AT_TARGET: number;
-    readonly CAPTURING_PHASE: number;
-    readonly NONE: number;
-};
+declare const GlobalCustomEvent: any;
```
jeremy-rifkin, line 2: This might not be correct, maybe this should actually implement `globalThis.CustomEvent`

jeremy-rifkin:
At the moment the types aren't right yet.

This fails for example:

```ts
const {
  window, document, customElements,
  HTMLElement,
  Event, CustomEvent
} = parseHTML("");
```

I can't figure out where properties like `window` are set for `Document`...

jeremy-rifkin:
At the moment there are a lot of types that are not right. Examples from the readme don't work.

esm/interface/node.js
WebReflection: they do, this is an interface like implementation

WebReflection: or you mean this should be the TSDoc change to make it happy? In such case, it's OK

WebReflection:
I'm afraid this PR needs some linting adjustment or the CI won't be happy ... and I care about 100% code coverage and happy CI here. Please have a look and yes, exceptions accepted for linters!

types/esm/index.d.ts
```diff
@@ -8,6 +8,6 @@ export { EventTarget } from "./interface/event-target.js";
 export { InputEvent } from "./interface/input-event.js";
 export { NodeList } from "./interface/node-list.js";
 export { NodeFilter } from "./interface/node-filter.js";
-export function parseHTML(html: any, globals?: any): Window & typeof globalThis;
+export function parseHTML(html: any, globals?: any): typeof globalThis;
```
jeremy-rifkin, line 11: not 100% confident about this typing

types/esm/interface/document.d.ts
```diff
@@ -1,20 +1,17 @@
...
- */
-export class Document extends NonElementParentNode implements globalThis.Document {
+export class Document extends NonElementParentNode {
     constructor(type: any);
     /**
-     * @type {globalThis.Document['defaultView']}
+     * @type {typeof globalThis}
      */
-    get defaultView(): Window & typeof globalThis;
+    get defaultView(): typeof globalThis;
```
jeremy-rifkin, line 6: not 100% confident about this type change

types/esm/interface/element.d.ts
```diff
@@ -20,8 +17,11 @@ export class Element extends ParentNode implements globalThis.Element {
     set tabIndex(arg: number);
     get tabIndex(): number;
     get innerText(): string;
-    set innerHTML(arg: any);
-    get innerHTML(): any;
+    /**
+     * @param {string} html
```
jeremy-rifkin, line 21: makes things more strongly typed

esm/interface/document.js
```diff
@@ -74,7 +71,7 @@ export class Document extends NonElementParentNode {
   }
 
   /**
-   * @type {globalThis.Document['defaultView']}
+   * @type {typeof globalThis & {document: globalThis.Document & Document}}
```
jeremy-rifkin, line 74: `& {document: globalThis.Document & Document}` allows `Document.toJSON`

jeremy-rifkin:
I think this PR is good to go

iMrDJAi:
> oh gosh ... all definition files [are automatically generated per each CI build step or locally](https://github.com/WebReflection/linkedom/blob/main/package.json#L19) before pushing to npm. Apologies I should've mentioned this as this PR is quite some effort if it was fixed manually per each generated file, but next release will re-create again all files and all changes will be lost.
> 
> The files you need to eventually change are in the esm folder, not in the types one, as that's all automation.

@WebReflection Hi, I want to contribute. Could you please explain what's the right way to include types? Are they generated from jsdoc comments?

esm/dom/string-map.js
```diff
@@ -25,9 +25,6 @@ const handler = {
   }
 };
 
-/**
- * @implements globalThis.DOMStringMap
```
WebReflection, line 29: any reason to remove this?

esm/dom/token-list.js
```diff
@@ -22,9 +22,6 @@ const update = ({[OWNER_ELEMENT]: ownerElement, value}) => {
     );
 };
 
-/**
- * @implements globalThis.DOMTokenList
```
WebReflection, line 26: any reason to remove this?

WebReflection:
@jeremy-rifkin 

> I think this PR is good to go

I just wonder once all "*pretending to implement*" classes are gone, if TS really helps out or it 'causes just confusion. There are things not really standard that might come up while using this project and I guess what are your thoughts there ... also, those two classes ... I think the implementation conforms with the standard? 🤔 are all implements needed to go?

WebReflection:
@iMrDJAi yes, this JS project uses [TSDoc](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html) and generates automatically types at build time.

jeremy-rifkin:
> I just wonder once all "_pretending to implement_" classes are gone, if TS really helps out or it 'causes just confusion. There are things not really standard that might come up while using this project and I guess what are your thoughts there ... also, those two classes ... I think the implementation conforms with the standard? 🤔 are all implements needed to go?

Thanks for taking a look! It may very well be these two classes do actually implement everything from their globalThis counterparts - I can re-introduce these `implements` annotations.

I think the one "pretending to implement" part remaining is whatever is returned by `Document.defaultView`. Alternatively to having it this function return a `Proxy` that's typed as `typeof globalThis & { document: globalThis.Document & Document; };` we could make a proper class / type for exactly what a linkedom document view is. This would address the issue of non-standard things coming up.

WebReflection:
this PR has conflicts ... please resolve those and I'd be happy to move forward, thanks!

WebReflection:
no answer in years ... closing.

Issue #174, opened 2022-10-21
Support for streams


iMrDJAi:
In order to reduce memory usage, and to improve performance, it's a smart move to provide support for streams, allowing to process *very* large inputs chunk by chunk. In this way, we don't have to load the whole thing in memory, and also, we may prevent `heap out of memory` errors.

htmlparser2 supports [streams](https://github.com/fb55/htmlparser2/wiki/Parser-options#streaming) out of the box, the implementation shouldn't be hard at all, but I have some concerns:

- Should we really make `parseFromString()` accept streams as `markupLanguage`? Wouldn't that break the DOM standard? Isn't a separate `parseFromStream()` more appropriate?

- Worried about Node.js streams support for cross environments support (browser/deno), including types.

Another thing: outputs as streams (https://github.com/WebReflection/linkedom/issues/98). This one would be trickier! a new method should be introduced to serve the purpose. The idea is we treat nodes as individual documents instead of appending them to the DOM, we keep processing their child nodes till we reach the close tag, then we push them into the stream flow (not necessarily a stream. It could be a callback, an EventEmitter), and we only do that to the ones that meet the criteria specified by user, for example we pass an array containing tag names or attributes that we need to catch (I can't see it possible with selectors but that would be neat!), and we ignore everything else. I feel like this is more like watching nodes than streaming them in my opinion.

Why not just using htmlparser2 directly instead? Well, we need processed output with DOM methods and properties, the ability to access child nodes, instead of the raw output that htmlparser2 returns, otherwise what's the point?

I'm actually willing to contribute, but I need your feedback first, you may have a better idea regarding that. Thank you.

WebReflection:
ehr .... differently from JSDOM this module parses pages with dozen MB of HTML without issues (it's part of the test cases) and it needs to link nodes to work for any string representation, [as explained in here](https://github.com/WebReflection/linkedom/issues/98#issuecomment-1015551489).

my feedback there is the same I'd give in here and I don't understand the use case for this or what you are trying to do.

partial content and selectors or methods or anything else would be broken anyway and there are ways to increment the heap, even if I've never needed it.

do you have a real-world case where this module breaks or a reason to crawl that page with LinkeDOM instead of using a real browser?

iMrDJAi:
[...]

Issue #98, opened 2022-01-18
Streaming HTML


devinellis:
I’m curious if it would be possible to use a stream, or loop through chunks of HTML into the parser? Looking at the source code I’m not sure if it’s possible, but the benefits for my use-case are:
- less memory usage (in theory?)
- Ability to drop nodes before adding them to the document (eg script tags)
- Stop parsing once it has met certain criteria

devinellis:
Sorry for the vagueness! I am crawling web sites and looking to use @mozilla/readability to extract the main content of the body.

The simplest solution is to wait for the whole response string and pass it into your parser, but sometimes all I need is the `<head>`, or the first N characters (or Nodes) of the body.

I could pre-process the HTML, but it seems ideal to do it all in one pass.

I like the interface of fb55/htmlparser2, and another benefit would be I’m not storing the whole document in memory twice: once as a string, and then as linkedon arrays.

Thanks for the quick reply!

WebReflection:
> I like the interface of fb55/htmlparser2

which is what we use in here too ... so ... just use that instead of this library? 'cause this library is for creating content, not for parsing streamed content, and I have no intention to create a dedicated parser for that, as I have honestly zero time but also this is not the project you should look after, because parsing streams means a constantly broken linked list of nodes, so it plays badly with the premisis of this lib.

Apologies if that's not the answer you were looking for, but I think you need to create what you need there, and htmlparser2 would be my hint too, as that's incremental, yet I have no idea if it works with streams, but if it does, your job is half-done.

Closing as no action will be taken, but feel free to ask more if needed.

PR #177, opened 2022-10-24
Streams support (pt. 2)

iMrDJAi:
Introducing `DOMStream`: an efficient way for parsing *very* large HTML/XML documents!

Usage:
```js
import {DOMStream} from 'linkedom'

const src = fs.createReadStream(join(__dirname, './big.xml')) // ~80 MB!
const domStream = new DOMStream('text/xml', (name, _attributes) => {
  // From htmlparser2's `onopentag`
  return name === 'hotelName'
})

src.pipe(domStream).on('document', doc => {
  // Magic!
  console.log(doc.documentElement.outerHTML)
})
```

Using `DOMParser`: Time: ~11mins, Memory: ~3GB (I got `heap out of memory` error but I was able to resume from JavaScript debug console).

image

Using `DOMStream` 🔥: Time: ~27s, Memory: ~13MB.

image


iMrDJAi:
> dunno mate, I really don't have any of your (folks) problems with LinkeDOOM because I don't care a bit about TS ... but sure ... as that's meant to solve the problem you mentioned, let's wait.

Fair enough. There is not much to do here anyway, `tsc` doesn't support `@property` yet (https://github.com/microsoft/TypeScript/issues/28730).

For future reference, I think users will have to explicitly specify the argument type of the 'document' event callback:

```ts
 const domStream = new DOMStream('text/xml', name => name === 'hotelName')
   .on('document', async (doc: XMLDocument) => {
     //
   })
```

esm/dom/stream.js
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * @type {{
+     *   document: MimeToDoc[MIME]
+     *   node: MimeToDoc[MIME]|Node
+     *   ownerSVGElement: SVGElement|undefined
+     *   rootNode: Node
+     * }[]}
+     */
+    this.stack = [];
+    this.init = this.init.bind(this);
```
WebReflection, line 44: why is this needed?

esm/dom/stream.js
```diff
@@ -0,0 +1,180 @@
...
+      document = new XMLDocument();
+    }
+    if (this.doctype) document.doctype = this.doctype;
+    this.stack.push({ document, node: document });
+  }
+
+  init ()  {
+    this.parserStream = new WritableStream({
+      // <!DOCTYPE ...>
+      onprocessinginstruction: (name, data) => {
```
WebReflection, line 64: could we share most of this code somehow instead of (looks like) repeating it all over? 🤔
I'm OK having a new file with repeated code for the time being, but I wonder why you didn't reuse methods already available elsewhere

WebReflection:
code coverage is not happy ... the goal here is 100% or nothing so far, it'd be great to keep it as such, thanks.

esm/dom/stream.js
WebReflection: P.S. it is very possible that sharing code will also help with code coverage

WebReflection:
@iMrDJAi you can `npm test` locally and see the table with the code coverage score ... if it's not 100% all rows and fields, is not good enough, if it is, great 🥳

WebReflection:
@iMrDJAi coverage down to 92% from 100% https://coveralls.io/builds/53677686 ... if there's some obvious case you think should not be covered you can use special comments around but as this is 100% new code I'd like to see it covered.

README.md
```diff
@@ -4,7 +4,7 @@
 
 <sup>**Social Media Photo by [JJ Ying](https://unsplash.com/@jjying) on [Unsplash](https://unsplash.com/)**</sup>
 
-### This is not a crawler!
+### This is ~~not a~~ better than a crawler! :3
```
WebReflection, line 7: I'd rather remove the sentence instead of making it pretentious as crawling is not the main goal of this project.

iMrDJAi:
> P.S. it is very possible that sharing code will also help with code coverage

@WebReflection I don't have to write more tests, I'll do what you've suggested earlier.

WebReflection:
will try to have a closer look locally soon, thanks for the refactoring though!

iMrDJAi:
> will try to have a closer look locally soon, thanks for the refactoring though!

No problem!

WebReflection:
a possible stopper https://github.com/WebReflection/linkedom/issues/180

this module works in JS Workers and if the stream implementation in third part libraries use something not Web compatible, I am not sure this should move forward ... it would rather be a dedicated parser usable only in NodeJS envs ... thoughts?

WebReflection:
the stream uses NodeJS specific modules, including emitter, so that pushing this change will break everything out there based on CouldFlare Workers or just Web JS Workers ... this is not going to happen, if my understanding is correct ... feel free to fork and make this module a NodeJS only project, as I have no interests in doing that, because there are already consumers using this module **not as a crawler** and that's the promise behind.

WebReflection:
beside apologizing for my last comment, this PR made me realize I need to improve testing via puppeteer or similar and be sure the module works in browsers, not just in servers.

iMrDJAi:
@WebReflection That's actually a concern I had at the beginning (https://github.com/WebReflection/linkedom/issues/174):

> Worried about Node.js streams support for cross environments (browser/deno), including types.

But there is definitely a solution for that. How about env check? polyfills? We may always include polyfills for node `stream` (streams are instances of events) and `string_decoder`.

And regarding CouldFlare Workers, it would work even without including polyfills: https://developers.cloudflare.com/workers/wrangler/configuration/#node-compatibility.

I'll try with it, it's all about installing a rollup plugin at the end.

iMrDJAi:
It needs to be tested

chr1se:
@iMrDJAi have you tried using [WebStream Api](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) to achieve those 🔥 awesome 🔥  benchmarks 

It is supported on all platforms:
[node](https://nodejs.org/api/webstreams.html#web-streams-api)
[deno-deploy](https://deno.com/deploy/docs/runtime-api#:~:text=Streams)
[cloudflare](https://developers.cloudflare.com/workers/learning/using-streams/)
[browser/webworkers](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API)

iMrDJAi:
@chr1se Is WebStream API considered a stable feature? Also, `htmlparser2` depends on NodeJS streams, so a new adapter is needed? Or we may just use both to achieve maximum compatibility (we can do a runtime environment check)?

WebReflection:
This module works in WebWorkers too … worker.js is meant to provide SSR for Service Workers PWA … polyfills are ok only if tiny and battle tested, otherwise are a burden to maintain I’ve no capacity… is a fork targeting NodeJS only with easy upstream sync really off the table? I don’t want to ruin this project scope which is, once again, not crawling.
Maybe a plugin system to bring your own parser could be a better solution?

iMrDJAi:
> I don’t want to ruin this project scope which is, once again, not crawling.

@WebReflection Not sure why do you assume that I'm interested in crawling, this is by no means my intention, just because the person who opened the first issue (https://github.com/WebReflection/linkedom/issues/98#issuecomment-1015542142) was trying to crawl doesn't mean that everyone else are doing the same. What I'm doing here is trying to achieve what `linkedom` has claimed to do:

> This structure also allows programs to avoid issues such as "Maximum call stack size exceeded" (basicHTML), or "JavaScript heap out of memory" crashes (JSDOM), thanks to its reduced usage of memory and zero stacks involved, hence scaling better from small to very big documents.

Sure it provided a huge improvement, but still, it doesn't scale well. ¯\\\_(ツ)_/¯  
The point is, this is under the scope of the library, according to README.

I'm aware that you're a busy man, and I respect that, however It's unlikely that anything I've added would force you to take an action in the future or cause complexity in any way, and as for polyfills, the build step is automatic, not sure what do you mean by "burden to maintain".

> Maybe a plugin system to bring your own parser could be a better solution?

No, that would be an overkill, most of the files are already classes, and classes are extendable?

WebReflection:
like I've said, if the worker.js file works in browsers out of the box and the resulting size is not increase by too many bytes, I'd be OK but it's a priority of this project to work **everywhere** and not inly in NodeJS ... bun, deno, **everywhere** like it does right now already. the polyfill part is in case they patch some breaking change and I get the blame, something unfortunately frequent in the npm world.

iMrDJAi:
@WebReflection Hello, sorry for the delay. Since polyfills add around 200 kb to the 300 kb `worker.js` (can be less if minified), are you okay with implementing Web Streams API instead? Otherwise, I'm okay with not merging and keeping my changes in my fork (unless you become interested or this feature gains more demand in the future, then I'll be happy to reopen the pull request again).

WebReflection:
What are the numbers once minified and gripped? But yeah, 3/4th of the current size for polyfills doesn’t look too compelling to me … but what if I publish a branch myself with the streaming option as linkedom-stream so that both maintaining it and synchronizing it would be under this repo? I’d be ok with that, or even a dedicated linkedom/stream export so no branch dance is needed, and most code gets shared by default

iMrDJAi:
> What are the numbers once minified and gzipped? But yeah, 3/4th of the current size for polyfills doesn’t look too compelling to me … but what if I publish a branch myself with the streaming option as linkedom-stream so that both maintaining it and synchronizing it would be under this repo? I’d be ok with that, or even a dedicated linkedom/stream export so no branch dance is needed, and most code gets shared by default

Actually, that's a good idea! A mono repo with multiple npm packages would work. Would you like to proceed?

iMrDJAi:
Closed since I got no confirmation from maintainer to proceed.

PR #185, opened 2022-11-14
Added `iframe` attributes

iam-medvedev:
Added following `iframe` attributes:

- [srcdoc](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-srcdoc) — A document to render in the iframe
- [name](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-name) — Name of nested navigable
- [allow](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-allow) — Permissions policy to be applied to the iframe's contents
- [allowFullscreen](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-allowfullscreen) — Whether to allow the iframe's contents to use `requestFullscreen()`
- [referrerPolicy](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-referrerpolicy) — Referrer policy for fetches initiated by the element
- [loading](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-loading) — Used when determining loading deferral

### DIFF
--- a/test/html/i-frame-element.js
+++ b/test/html/i-frame-element.js
@@ -13,10 +13,10 @@ assert(iframe.src, './test.html', 'Issue #82 - <iframe>.src');
   const { document } = parseHTML("<html><body><iframe></iframe></body></html>");
   const iframe = document.body.querySelector("iframe");
 
-  iframe.srcdoc = `<html><span>Test</span></html>`;
+  iframe.srcdoc = `<html><span style="color: red">Test</span></html>`;
   assert(
     document.body.innerHTML,
-    `<iframe srcdoc="<html><span>Test</span></html>"></iframe>`
+    `<iframe srcdoc="<html><span style=&quot;color: red&quot;>Test</span></html>"></iframe>`
   );
 }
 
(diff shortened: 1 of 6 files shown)


test/html/i-frame-element.js
```diff
@@ -7,3 +7,55 @@ const {document} = parseHTML('<html><iframe src="./test.html"></html>');
...
+
+
+{
+  const { document } = parseHTML("<html><body><iframe></iframe></body></html>");
+  const iframe = document.body.querySelector("iframe");
+
+  iframe.srcdoc = `<html><span>Test</span></html>`;
+  assert(
+    document.body.innerHTML,
+    `<iframe srcdoc="<html><span>Test</span></html>"></iframe>`
```
WebReflection, line 19: if this pass, there's a missing escape in the process

iam-medvedev: According to [spec](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-srcdoc) only quotes should be escaped. 

For example (browser):
```js
const iframe = document.createElement('iframe');
iframe.srcdoc = `<html><span style="color: red">Test</span></html>`;
console.log(iframe.outerHTML); // <iframe srcdoc="<html><span style=&quot;color: red&quot;>Test</span></html>"></iframe>
```
I [updated](https://github.com/WebReflection/linkedom/pull/185/commits/ed2b4efd430aadb7d4ca098194f0f667881575fd) the test and it seems to work correctly.

PR #190, opened 2023-02-12
feat/namespaceURI

kyr0:
Hi Andrea,

this PR implements `Element.namespaceURI`, as far as I understood, close to WHATWG spec. requirements:

- `namespaceURI` is a getter that returns the namespace string that has been provided when calling `createElementNS`
- when an `SVGElement` is created, the SVG namespace is passed to the base class
- all `SVGElement` instances now automatically come with the `namespaceURI` accessible on them
- the code follows the common patterns in this repo and therefore introduces a new `Symbol` that is used to store the `namespaceURI` on the instance of the Element
- the `option` argument in `createElementNS` has been removed, it was unused 
- because `createElement` could potentially take an `options` argument in the future (as by the spec), I don't call `createElement` anymore from within `createElementNS` as this could lead to future breaking changes when `options` should ever be implemented -- therefore the internal API (Element class) takes another constructor parameter now

All tests are green, 100% code coverage. Added a few more tests to make sure the change translates well down the tree for the SVG usecases.

Double-checked with local npm linking -- all my tests with the isomorphic DOM renderer are green now :)

Thanks and best
Aron


WebReflection:
before I look into this (which I will right after) I wonder why you dropped something that was there to be future proof (updates) ... as the `options` is [standard API signature](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElementNS#syntax) ... any particular reason?

esm/interface/document.js
```diff
@@ -269,10 +269,10 @@ export class Document extends NonElementParentNode {
   createAttributeNS(_, name) {
     return this.createAttribute(name);
   }
-  createElementNS(nsp, localName, options) {
+  createElementNS(nsp, localName) {
     return nsp === SVG_NAMESPACE ?
             new SVGElement(this, localName, null) :
-            this.createElement(localName, options);
```
WebReflection, line 275: options is used for builtin extends ... I think this breaks builtin extend when created via `createElementNS` but I need to check if browsers also do that

esm/interface/element.js
```diff
@@ -75,11 +75,19 @@ const isVoid = ({localName, ownerDocument}) => {
  * @implements globalThis.Element
  */
 export class Element extends ParentNode {
-  constructor(ownerDocument, localName) {
+  constructor(ownerDocument, localName, nsp) {
     super(ownerDocument, localName, ELEMENT_NODE);
     this[CLASS_LIST] = null;
     this[DATASET] = null;
     this[STYLE] = null;
+    this[NS] = nsp
```
WebReflection, line 83: this is an implicit property of the element ... as this project only accepts SVG or non SVG elements, an `instanceof SVGElement` seems to be enough to return via lazy accessor the `nsp` ... adding extra fields to the class looks like a road to worse performance, so lazyness here would be much better, imho. What is the reason to attach something 90% of the use cases not needed?

esm/interface/element.js
```diff
@@ -75,11 +75,19 @@ const isVoid = ({localName, ownerDocument}) => {
...
+  }
+
+  /**
+   * @returns {String}
+   */
+  get namespaceURI() {
+    return this[NS]
```
WebReflection, line 90: as written before, this is where `instanceof` check seems ot be more appropriate for *SVG* elements ... I myself never needed any of this in the wild, so let's make it lazy, please?

esm/svg/element.js
```diff
@@ -17,7 +18,7 @@ const handler = {
  */
 export class SVGElement extends Element {
   constructor(ownerDocument, localName, ownerSVGElement = null) {
-    super(ownerDocument, localName);
+    super(ownerDocument, localName, SVG_NAMESPACE);
```
WebReflection, line 21: every `SVGElement` element has implicitly `SVG_NAMESPACE` as `namespaceURI` ... I feel like these elements are the only one that should have something meaningful in here.

test/html/document.js
```diff
@@ -88,3 +88,10 @@ const issue187 = (new DOMParser).parseFromString(`<!DOCTYPE html><html><head></h
 issue187.body.innerHTML = '<span></span>';
 let span = issue187.body.firstElementChild;
 assert(span.ownerDocument.body, issue187.body);
+
+// namespaces
+const issue189Div = document.createElementNS("http://www.w3.org/1999/xhtml", "DIV")
+assert(issue189Div.namespaceURI, 'http://www.w3.org/1999/xhtml', '<div> element, when created with createElementNS() must keep its namespace')
```
WebReflection, line 94: if this is the reason for the whole dance then I'd like to introduce an `XHTML` variant among `XML` and `SVG` and move the namespace as static getter through the prototype in all cases ... this makes the code identical as it is and just enrich edge cases instead of penalizing with an extra field all elements ... what do you think?

WebReflection: Actually ... never mind, this *is* the namespaceURI for HTML: https://developer.mozilla.org/en-US/docs/Web/API/Document/createElementNS#important_namespace_uris

And here I've done this test:

```js
customElements.define('a-li', class extends HTMLLIElement{}, {extends: 'li'});
document.createElement('li', {is: 'a-li'}) instanceof customElements.get('a-li'); // true
document.createElementNS('http://www.w3.org/1999/xhtml', 'li', {is: 'a-li'}) instanceof customElements.get('a-li'); // also true
```

So this PR breaks custom elements builtin extends on `createElementNS` with valid HTML namespace, hence it's a no-go to me as it is. Let's move `namespaceURI` in respective class prototypes instead, so that no performance regression can really happen, and nothing would break with builtin extends.

WebReflection: P.S. dare I say this test should be part of the coverage too, so these issues would be clear out of the box.

esm/interface/document.js
WebReflection: checked here: I am afraid this is not going to lad as is https://github.com/WebReflection/linkedom/pull/190#discussion_r1103862030

esm/svg/element.js
WebReflection: Amend, I am OK to have either lazy results or specific `get namespaceURI` for SVG elements that override the default for HTML one.

WebReflection:
I went ahead and fixed this issue with the most sensible and minimal change, I hope that's OK.

PR #192, opened 2023-03-04
Add `meta` attributes

jarusll:
Added the following [meta](https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-media) attributes
- name - The metadata key
- content - The metadata value
- http-equiv - Special flags for the page
- charset - The character encoding
- media - Preffered stylesheet according to media queries

### DIFF
--- a/.github/workflows/node.build.js.yml
+++ b/.github/workflows/node.build.js.yml
@@ -1,0 +1,30 @@
+# This workflow will do a clean installation of node dependencies, cache/restore them, build the source code and run tests across different versions of node
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs
+
+name: Node.js CI
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  build:
+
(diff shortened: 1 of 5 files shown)


WebReflection:
It's up and running 🥳

jarusll:
Am I  supposed to push `worker.js` as well? 
Also you might want to delete the github action `[node.build.js.yml](https://github.com/WebReflection/linkedom/blob/main/.github/workflows/node.build.js.yml)`, I pushed it without realizing there was already one present.

jarusll:
If [node.build.js.yml](https://github.com/WebReflection/linkedom/blob/main/.github/workflows/node.build.js.yml) is not deleted by the next PR, I will include it in the PR. Thanks and please let me know if I am supposed to push `worker.js` or not.

WebReflection:
worker comes out of hte build which happens anyway before I push versions from main ... in theory all PRs need is to change the *./esm/* and the *./test/* entry and everything else is handled by the pipe ... I will remove that .yml file then.

WebReflection:
.yml removed ... maybe I could take a chance to update mine with more versions and a v3 checkout ... right now that's not a priority neither necessary, I hope that's OK.

PR #195, opened 2023-03-09
Ensuring /cached and /worker entrypoints are recognized by TypeScript

### LINKED ISSUE #194: Ensuring the /cached export is detected by TypeScript
It looks like the `/cached` export is not detected by TS:

```ts
import * as A from 'linkedom';
import * as B from 'linkedom/cached';
import * as C from 'linkedom/worker';
```

Screen Shot 2023-03-09 at 18 34 45

I'm not exactly sure why, maybe a "types" property is required [here](https://github.com/WebReflection/linkedom/blob/e22febd73d226dd9310a9e530679f08fbe2491c7/package.json#L51).

fabiospampinato:
I remembered that I stumbled on this same issue with [Voby](https://github.com/vobyjs/voby/blob/d9712d8a9dc770891522591b2548c756c7ca8acd/package.json#L24-L33), and after some major headaches the solution was that there just was no reasonable way to provide types for submodules, the only way to do it is by abusing the `"typesVersions"` property. This is maybe something that should get addressed in TS v5.

In any case this PR closes #194.

I'm reusing the types for the default export for the worker export, since I don't see more specific ones for that, if /worker exports a different interface then different types should be generated I guess.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,16 @@
     "./package.json": "./package.json",
     "./worker": "./worker.js"
   },
+  "typesVersions": {
+    "*": {
+      "cached": [
+        "./types/esm/index.d.ts"
+      ],
+      "worker": [
+        "./types/esm/index.d.ts"
+      ]
+    }
+  },
   "dependencies": {
     "css-select": "^5.1.0",
     "cssom": "^0.5.0",


WebReflection:
it's up, thanks 👍

PR #207, opened 2023-06-04
fix: mutation observer callback during parse

luwes:
thanks for this great piece of software! really awesome work!

this fix seemed to be needed to have a more similar behavior as how the browser behaves but I'm not 100% sure it should go there.

if it makes sense let me know.


WebReflection:
please rebase from main and resolve conflicts + please add a test ... AFAIK MO works already as expected so I am not even sure this is needed.

luwes:
Thanks. I'll close it until it comes up for someone else. 

I keep a branch up to date with a few fixes I need 
https://github.com/luwes/linkedom/commits/wesc

PR #209, opened 2023-06-10
Improve <slot> element support

### LINKED ISSUE #131: HTMLSlotElement.assignedNodes()
Currently `HTMLSlotElement.assignedNodes()` doesn't seem to be supported, I have a couple of components that I'd like to SSR that make use of this API. Do you think `assignedNodes()` could at some point be supported?

luwes:
Fixes #131

- Adds the `name` property to the slot element
- Adds `assignedNodes()` and `assignedElements()` methods
- Adds the `slot` property to the Element class

### DIFF
--- a/esm/html/element.js
+++ b/esm/html/element.js
@@ -98,8 +98,6 @@ export class HTMLElement extends Element {
   set lang(value) { stringAttribute.set(this, 'lang', value); }
   get title() { return stringAttribute.get(this, 'title'); }
   set title(value) { stringAttribute.set(this, 'title', value); }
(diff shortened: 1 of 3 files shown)


WebReflection:
I'm OK with this MR but there's no test whatsoever ... it's OK to skip testing repeated, already tested, logic such as utilities, but here we have a whole new functionality and there's gonna be something covering that functionality.

At least that's how every other MR got merged, thanks!

WebReflection:
OK, I am going to merge this but I'll file an issue about adding proper tests to it ... which I just did: https://github.com/WebReflection/linkedom/issues/232

PR #213, opened 2023-07-18
fix: getRootNode() bug

luwes:
the `<html>` element is [HTMLHtmlElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLHtmlElement)

`getRootNode()` returns 

An object inheriting from Node. This will differ in exact form depending on where you called getRootNode(); for example:

- Calling it on an element inside a standard web page will return an HTMLDocument object representing the entire page (or <iframe>).
- Calling it on an element inside a shadow DOM will return the associated ShadowRoot.

---

I ran in to this bug when calling getElementById on the root node but the method didn't exist with Linkedom.
Code here https://github.com/muxinc/media-chrome/blob/main/src/js/media-theme-element.js#L177-L178

### DIFF
--- a/cjs/interface/node.js
+++ b/cjs/interface/node.js
@@ -188,11 +188,16 @@ class Node extends EventTarget {
     return this.parentNode;
   }
 
+  /**
+   * Calling it on an element inside a standard web page will return an HTMLDocument object representing the entire page (or <iframe>).
+   * Calling it on an element inside a shadow DOM will return the associated ShadowRoot.
+   * @return {ShadowRoot | HTMLDocument}
+   */
   getRootNode() {
     let root = this;
     while (root.parentNode)
       root = root.parentNode;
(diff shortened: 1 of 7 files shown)


WebReflection:
This breaks the test which was already showing the `getRoot()` returns the `documentElement` so it's not clear what this fixes or why it's needed.

luwes:
^^ updated

WebReflection:
OK, thanks! this looks like an unintentional bug indeed ... however, code-coverage isn't super happy now ... mind trying to understand what's going on or why something is not covered?

```sh
npx c8 report --reporter=html
```

this should give you a `coverage/index.html` with all stats per file so you can see which one is not covered anymore or why.

WebReflection:
it was passing through the `||` before, but it works as expected so we're good now ... it's up and running.

Issue #210, opened 2023-06-23
Support getInnerHtml() for SSR of declarative shadow DOM


ttraenkler:
linkedom would be a great fit for lightweight SSR of web components, missing only ```Element.getInnerHTML()```. [Here](https://gist.github.com/developit/54f3e3d1ce9ed0e5a171044edcd0784f) is a polyfill that might serve as a starting point for an implementation.

For server side rendering of web components, [declarative shadow DOM](https://developer.chrome.com/articles/declarative-shadow-dom/) has recently been standardized and [shipped](https://caniuse.com/?search=declarative%20shadow%20dom) in all major browsers except Firefox, [Mozilla's position being positive](https://github.com/mozilla/standards-positions/issues/335) on this feature as well.

Similar to ```Element.innerHTML```, the new function ```Element.getInnerHTML()``` will [serialize DOM elements including the shadowroots of custom elements](https://github.com/mfreed7/declarative-shadow-dom#serialization) which are skipped by innerHTML. In essence, wherever a ```<custom-element>``` is found in the DOM, the corresponding template will be inlined with ```<template shadowrootmode="open">``` (note this was recently renamed from shadowroot="open" which the above mentioned polyfill still uses).

Instead of
```ts
element.innerHTML
```
producing
```html
   <custom-element>
      <h2>light content</h2>
    </custom-element>
```
the new function
```ts
element.getInnerHTML({includeShadowRoots: true})
```
will include the shadowroot
```html
   <custom-element>
      <template shadowrootmode="open">
        <p id="a">shadow a</p>
        <p id="b">shadow b</p>
        <p id="c">shadow c</p>
        <slot></slot>
      </template>
      <h2>light content</h2>
    </custom-element>
```

WebReflection:
where would the content of that shadowroot be taken from?

WebReflection:
also, what happens with mode `closed` ?

ttraenkler:
The polyfill I linked in the OP contains code to deal with open and closed shadowroots. 

> where would the content of that shadowroot be taken from?

For open shadowroots simply from the element's ```this.shadowRoot```.

> also, what happens with mode `closed` ?

For closed you pass the closed shadowroots to the getInnerHTML function explicitly as an array. Here is a description from the champion of the proposal that explains it: 

_Closed shadow roots
To allow serialization of closed shadow roots, an additional option will be added:

[...]

Issue #178, opened 2022-10-24
Calling this.attachShadow(...) immediately adds all childNodes to the shadowDom


chr1se:
The expected behaviour is for childNodes to only be manually added to the shadowdom (root.append, etc) or automatically via [declarative shadowdom](https://web.dev/declarative-shadow-dom/) 

https://github.com/WebReflection/linkedom/blob/6a8b083e1b0e71168e3bef2657b98bdad492992b/worker.js#L7515
removing the line above fixed it.

Usecase:  I am doing SSR of pages with a bunch of webcomponents that uses the declarative shadowdom syntax (DeclarativeSD are only supported by chromium based browsers, but i use a polyfill for firefox & safari). childNodes were disappearing after calling this.attachShadow(...) in the constructor of the custom element definition. 

thx

WebReflection:
> webcomponents that uses the declarative shadowdom syntax

not really ... I want DSD fully available in here once it's fully standard and adopted 😊

WebReflection:
to clarify, Shadow DOM in here is not really a goal or well supported ... slots don't exist, and so on ... I am waiting for the right time to do it right so an intermediate "still broken" solution wouldn't work for the time being, or a major update.

chr1se:
thx for response, 2 things: 

firstly, are you saying that some future date when DSD is fully standardised across all browsers (& you have free time :wink:), you will then fully implement shadowDom, DSD & slots in linkedDom, but until then you will leave it in its current state as it is ([with line 7515](https://github.com/WebReflection/linkedom/blob/6a8b083e1b0e71168e3bef2657b98bdad492992b/worker.js#L7515) below)?


```javascript
  attachShadow(init) {
    if (shadowRoots.has(this))
      throw new Error('operation not supported');
    // TODO: shadowRoot should be likely a specialized class that extends DocumentFragment
    //       but until DSD is out, I am not sure I should spend time on this.
    const shadowRoot = new ShadowRoot$1(this);
    shadowRoot.append(...this.childNodes); //<===========| this line is the issue |========<
    shadowRoots.set(this, {
      mode: init.mode,
      shadowRoot
    });
    return shadowRoot;
  }
```

[...]

PR #214, opened 2023-07-20
adds support for getInnerHTML (declarative shadow DOM)

ttraenkler:
Adds support for `getInnerHTML` fixing https://github.com/WebReflection/linkedom/issues/210.

@WebReflection: Since this is my first larger contribution, I would appreciate early feedback  if you need anything to be changed. I have tried to adopt the coding style used in the project and added minimal unit tests.

As we [discussed](https://github.com/WebReflection/linkedom/issues/178#issuecomment-1605403957) for this first implementation I decided to adapt an [existing polyfill](https://gist.github.com/developit/54f3e3d1ce9ed0e5a171044edcd0784f) from @developit. 

It uses innerHTML to serialize the shadow root which could optimized away, but might require more complete support for shadow DOM so I am hesitant to do this now before shadow DOM is properly supported in linkedom. The polyfill already improves on the current state of no support for _declarative_ shadow DOM enough that I would suggest to leave optimization to a separate issue as it might require work on broader support for shadow DOM and would block this issue until that has landed. 

This way we can start broader testing earlier in the spirit of [make it work, make it right, make it fast](https://wiki.c2.com/?MakeItWorkMakeItRightMakeItFast) and we can identify which shadow DOM features are missing or need to be fixed (like #178 which I have stumbled upon and fixed along the way). 

As I found in the comments I made ShadowRoot inherit from DocumentFragment.


WebReflection:
Thanks @ttraenkler, I'll try to have a look ASAP but I won't likely do that until coverage is back to 100%. If there is some path *really* hard to reason about (for coverage/edge cases) then feel free to add a c8 comment to ignore that step with some explanation about *why* too, thanks!

ttraenkler:
@WebReflection Yes, I just noticed and fixed this.

WebReflection:
This MR is still in a **draft** state and it has merge conflicts ... please make it review ready after rebasing from main, thanks!

ttraenkler:
As discussed I clarified with the champion of the getInnerHTML proposal that this is not yet standardized and likely never will due to lack of consensus. Therefore closing. Thanks anyways for considering!

PR #222, opened 2023-10-07
Add deep argument to the type of cloneNode()

tai2:
`Node`'s `cloneNode()` should take the `deep` argument in type definition. Otherwise, other derived classes like `Element` also cannot take that parameter in TS.


WebReflection:
to be honest, I am not sure this is a TS shenanigan but `cloneNode` already takes an optional argument ... I'll merge it, let's hope it will stay around ... if it doesn't, I am afraid I need to change the signature in the code, not in its `.d.ts` definition.

Issue #245, opened 2023-11-01
[Feature] document.createElement() API should create matching derived type of html element


Havunen:
Currently when using `document.createElement("div")` it creates instance of `Element` class where tag name equals "div".
However Linkedom has all these https://github.com/WebReflection/linkedom/tree/main/esm/html derived types defined in the library but they are not actually used when creating instances of html nodes through JS API. 

This feature should make following code pass
```js
var dom = document.createElement("div");

expect(dom instanceof HTMLDivElement).toEqual(true);
```

Havunen:
I changed type of this issue to BUG because I noticed it works just like for some tags but not for all

Issue #247, opened 2023-11-01
Anyone willing to help this project forward?


WebReflection:
I think this project works awesomely for anything I need it to do, but I am afraid I am a full time contractor and, before that, a full time dad of a 21 months old child that leaves me pretty much zero extra time to deal with bugs, PRs reviews, publish, release related things, and so on ... I also don't want to abandon this project because I think it's pretty cool and robust enough for SSR related tasks, but it's clear that many keep trying to use this to test things or pretend this project is like JSDOM and behave like a browser ... I have two options:

  * make this project "locked" for its own users (no issues, no PRs, nothing, I just work on it when and if I have time)
  * find other maintainers that just stick with the contribution docs and don't waste time with stuff that ain't needed (like change linter rules, make everything `.ts` or anything that is not moving really the project further, just churn no current issue needs)

As summary: is anyone willing to jump in? All I can do is to guarantee that any question around any current decision can be answered, while maintainers can help this project forward while I do completely different things in my daily job which has nothing to do with LinkeDOM improvements ... what do you think? Any volunteer?

Havunen:
I could try to help, but my time is also very limited.

I recently started exploring different approaches to speed up jest tests and one of the reasons why it is slow is JsDOM. JsDOM tries to fulfill the html specification, but what I need is javascript API for HTML operations, not the full API but the basics should work so it could be used as testing environment. The difficult stuff could be mocked out or anything which would make it slower. 

I initially started building my own library and then I found this project which was very useful, great project!

Here is the jest plugin I'm working on:
https://github.com/Havunen/jest-environment-linkedom

pkra:
While I'm more of a bystander, I thought that might be useful to write something.

I've been using linkedom as my daily driver for well over a year now, doing SSR-ish things, replacing jsdom. I ran into 4 bugs in that time, 1 of which was 3rd party. You kindly fixed them which I'm very grateful for (and even the 3rd party one was eventually fixed by someone).

I really like how clearly you communicate your approach in the project. I have always been able to do what I needed to do using linkedom. In fact, if I ran into issues, resolving it usually improved my code.

WebReflection:
@Havunen 

> Here is the jest plugin I'm working on:

that's the thing ... I don't have time for this project right now, I have even less time to "check other projects out".

You landed a few PRs so I'll try this evening or over the weekend to have a look and if I like that, I'd be happy to add you as contributor ... does this sound reasonable to you?

Havunen:
[...]

PR #248, opened 2023-11-01
Added all tags to html class map to support all derived tag type through createElement API

Havunen:
Fixes https://github.com/WebReflection/linkedom/issues/245

### DIFF
--- a/cjs/html/area-element.js
+++ b/cjs/html/area-element.js
@@ -1,12 +1,17 @@
 'use strict';
 const {HTMLElement} = require('./element.js');
+const {registerHTMLClass} = require("../shared/register-html-class.js");
+
+const tagName = 'area';
 
 /**
  * @implements globalThis.HTMLAreaElement
  */
 class HTMLAreaElement extends HTMLElement {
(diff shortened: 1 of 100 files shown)


WebReflection:
ehm ... **no** ... if you read the how to contribute the classes you see are placeholders.

If no special meaning or type is needed, this is not going to happen because it's crytical on the performance path.

If you ned to check `instanceof` it means you are using *LinkeDOM* for testing purpose and I am tired to state this is not the goal of the library so while I really do appreciate your will to contribute, this is **not** what this library needs + I think your PR is not doing what you think it's doing.

Special elements pollute the parsing and bootstrap all over the place ... anything that behaves like a div is good enough by being just an Element ... everything else, unless really needed, doesn't need to be a special case.

Select and a few others are examples you can see how these get registered, other classes are here to simplify augmenting their presence and nature on the dom, not to have 1:1 instanceof operations.

I am afraid this is not landing as it is.

WebReflection:
a better answer in here: https://github.com/WebReflection/linkedom/issues/247#issuecomment-1904087838

PR #254, opened 2023-12-11
Implementation of `customElements.getName()`

mash-graz:
Fixes: #253

### DIFF
--- a/esm/interface/custom-element-registry.js
+++ b/esm/interface/custom-element-registry.js
@@ -195,7 +195,7 @@ export class CustomElementRegistry {
    * @param {Function} Class **Class** of custom element
    * @returns {string?} found tag name or null
    */
-    getName(Class){
+    getName(Class) {
       if (Classes.has(Class)) {
         const { localName } = Classes.get(Class);
         return localName;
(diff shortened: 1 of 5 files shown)


esm/interface/custom-element-registry.js
```diff
@@ -190,4 +190,16 @@ export class CustomElementRegistry {
     const info = this.registry.get(localName);
     return info && info.Class;
   }
+
+  /**
+   * @param {Function} Class **Class** of custom element
+   * @returns {string?} found tag name or null
+   */
+    getName(Class){
```
WebReflection, line 198: some indentation issue here ... one less space for `getName` and the rest of the code, otherwise this looks good to me.

mash-graz: Should be fixed now!

Thanks for your fast response.

esm/interface/custom-element-registry.js
```diff
@@ -195,7 +195,7 @@ export class CustomElementRegistry {
    * @param {Function} Class **Class** of custom element
    * @returns {string?} found tag name or null
    */
-    getName(Class){
+    getName(Class) {
```
WebReflection, line 198: this is also good but I wasn't clear ... see indentation compared to get(...)

mash-graz: O.k. -- I finally got it! :)

Indentation should be fixed now.

Issue #260, opened 2024-03-05
[BUG] `innerHTML` on elements parsed from XMLDocument doesn't escape characters as XML entities inside attribute values


omril1:
Repro:
```js
var linkedom = require('linkedom')
var parser = new linkedom.DOMParser()
var doc = parser.parseFromString(`<hierarchy><android.view.View content-desc="text3&amp;more"/></hierarchy>`, 'text/xml')
console.log(doc.documentElement.innerHTML) // '<android.view.View content-desc="text3&more" />'
```
Using the same code on a DOMParser from a browser or JSDOM gives `<android.view.View content-desc="text3&amp;more"/>`

WebReflection:
It's in: https://github.com/WebReflection/linkedom/pull/261

PR #261, opened 2024-03-05
Fix issue #260 escape for attributes in `XMLDocument`s

omril1:


esm/interface/element.js
```diff
@@ -228,6 +229,9 @@ export class Element extends ParentNode {
     if (name === 'class')
       return this.className;
     const attribute = this.getAttributeNode(name);
+    if (this.ownerDocument[MIME].docType.startsWith('<?xml')) {
```
omril1, line 232: I'm not sure what the ideal way to check if the element is part of an XML document, having it like this for now to be corrected

omril1: Another way I see might be `this.ownerDocument[MIME].docType === Mime['application/xml'].docType` (`this.ownerDocument[MIME] === Mime['application/xml']` doesn't seem to work)

esm/interface/element.js
```diff
@@ -228,6 +229,9 @@ export class Element extends ParentNode {
     if (name === 'class')
       return this.className;
     const attribute = this.getAttributeNode(name);
+    if (!ignoreCase(this)) {
```
WebReflection, line 232: this check is useless if `attribute` is `null` or `undefined` ... can we put it behind as suggested in my previous comment?

```js
return attribute && (ignoreCase(this) ? attribute.value : escape(attribute.value));
```

Thanks!

test/interface/element.js
```diff
@@ -17,4 +17,11 @@ div.innerHTML = '<span></span>'
...
-assert(div.namespaceURI, 'http://www.w3.org/1999/xhtml');
\ No newline at end of file
+assert(div.namespaceURI, 'http://www.w3.org/1999/xhtml');
+
+const parser = new DOMParser();
+const doc = parser.parseFromString(`<hierarchy><android.view.View content-desc="text3&amp;more"/></hierarchy>`, 'text/xml').documentElement;
+
+assert(doc.firstChild.getAttribute('content-desc'), 'text3&amp;more');
+assert(doc.firstChild.outerHTML, '<android.view.View content-desc="text3&amp;more" />');
+assert(doc.innerHTML, '<android.view.View content-desc="text3&amp;more" />');
```
WebReflection, line 27: I think we should add another test with the same content but for `text/html` case otherwise code-coverage is not happy

omril1: I think `emptyAttributes` shouldn't apply to XML so I changed it a bit

cjs/interface/attr.js
```diff
@@ -42,8 +43,11 @@ class Attr extends Node {
 
   toString() {
     const {name, [VALUE]: value} = this;
-    return emptyAttributes.has(name) && !value && ignoreCase(this) ?
-      name : `${name}="${value.replace(QUOTE, '&quot;')}"`;
+    if (!value) {
```
WebReflection, line 46: this changes the outcome or checks too much for all attributes without a value ... having `emptyAttributes.has(name)` as first guard for the common HTML case is desired and XML attributes are covered by the `ignoreCase(this)` check ... I would put that guard back ahead of the rest to avoid perf regression (too bad I don't have automated checks for that).

WebReflection: XML is the edge/non-common case, HTML is the primary use case to date so ... thanks!

WebReflection:
we're almost there ... code-coverage still not happy ... mind trying to make it happy? there are 6 cases:

  * it's HTML and
    * the attribute is an empty known one and its value is not empty
    * the attribute is an empty known one and its value is empty
    * the attribute is not empty
  * it's XML and
    * the attribute is an empty known one and its value is not empty
    * the attribute is an empty known one and its value is empty
    * the attribute is not empty

these should cover these changes and pass through all logic branches, if I haven't missed any other case, thanks!

PR #266, opened 2024-03-22
Fixed: inspecting document.title deleted it as a side-effect (when cached)

philipp-classen:
Minimal reproducible example with linkedom 0.16.10:

```
import { parseHTML } from 'linkedom/cached';

const html = '<!DOCTYPE html><html><head><title>Test</title></head><body></body></html>';
const { document } = parseHTML(html);
console.log('Title:', document.title); // Title: Test
console.log('Title:', document.title); // Title:          <-- missing (title removed as a side-effect)
```

### DIFF
--- a/esm/html/document.js
+++ b/esm/html/document.js
@@ -77,13 +77,13 @@ export class HTMLDocument extends Document {
    */
   get title() {
     const {head} = this;
-    let title = head.getElementsByTagName('title').shift();
+    let title = head.getElementsByTagName('title')[0];
     return title ? title.textContent : '';
   }
 
   set title(textContent) {
     const {head} = this;
-    let title = head.getElementsByTagName('title').shift();
+    let title = head.getElementsByTagName('title')[0];
     if (title)
       title.textContent = textContent;
     else {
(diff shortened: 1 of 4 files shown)


esm/html/document.js
```diff
@@ -77,13 +77,13 @@ export class HTMLDocument extends Document {
    */
   get title() {
     const {head} = this;
-    let title = head.getElementsByTagName('title').shift();
+    let title = head.getElementsByTagName('title')[0];
```
WebReflection, line 80: I think these days I would go for `.at(0`) instead ... so this could be:

```js
get title() {
  const {head} = this;
  return head.getElementsByTagName('title').at(0)?.textContent || '';
}
```

esm/html/document.js
```diff
@@ -77,8 +77,7 @@ export class HTMLDocument extends Document {
    */
   get title() {
     const {head} = this;
-    let title = head.getElementsByTagName('title')[0];
-    return title ? title.textContent : '';
+    return head.getElementsByTagName('title')[0]?.textContent || '';
```
WebReflection, line 80: `.at(0)` ... not `[0]` ... `[0]` bails out performance if no `title` exists, `.at(0)` doesn't suffer this issue.

WebReflection:
Thanks! It's up and running. If you had an issue please be sure that gets closed too 👋

PR #275, opened 2024-06-04
Allow optional no character escape on getAttribute

stefandokic-tricentis:


WebReflection:
https://developer.mozilla.org/en-US/docs/Web/API/element/getAttribute

I am not implementing non standard signatures and you can `unescape` through the same `html-escaper` module linkedom uses when you don't want escaped content ... not sure this is a good thing to land, or better, it won't land as deviation from the standard signature.

stefandokic-tricentis:
> https://developer.mozilla.org/en-US/docs/Web/API/element/getAttribute
> 
> I am not implementing non standard signatures and you can `unescape` through the same `html-escaper` module linkedom uses when you don't want escaped content ... not sure this is a good thing to land, or better, it won't land as deviation from the standard signature.

Thanks for clarifying.

PR #278, opened 2024-06-09
+ DocumentFragment.textContent

### LINKED ISSUE #231: null textContent value for DocumentFragment
The `.textContent` attribute of a `DocumentFragment`—such as the one returned by a `<template>`’s `.content` attribute—is `null`.

```js
import { parseHTML } from "linkedom";

const { document } = parseHTML("");

const template = document.createElement("template");
template.innerHTML = "foo <span>bar</span>";

// in linkedom this is "foo bar", in browsers it’s an empty string
console.log(template.innerText);

const documentFragment = template.content;

// in linkedom this is null, in browsers it’s "foo bar"
console.log(documentFragment.textContent);
```

iacore:
fix #231

this is certainly a fix. i don't know how efficient it is.


WebReflection:
There is no test, please add one, thanks.

esm/interface/document-fragment.js
```diff
@@ -8,4 +9,15 @@ export class DocumentFragment extends NonElementParentNode {
   constructor(ownerDocument) {
     super(ownerDocument, '#document-fragment', DOCUMENT_FRAGMENT_NODE);
   }
+
+  get textContent() {
+    const text = [];
+    let {[NEXT]: next, [END]: end} = this;
+    while (next !== end) {
+      if (next.nodeType === TEXT_NODE)
+        text.push(next.textContent);
```
WebReflection, line 18: not sure if `node.data` is what you want but I think should do it too

iacore: `node.data`?

what i want is to fix https://github.com/WebReflection/linkedom/issues/231

iacore:
[the plot thickens](https://github.com/lumeland/lume/discussions/611#discussioncomment-9720365)

i may end up adding more features.

esm/interface/document-fragment.js
WebReflection: I meant `next.data` which is a direct getter ... `textContent` passes through `data` so is a tiny bit slower https://github.com/WebReflection/linkedom/blob/main/esm/interface/character-data.js#L41-L51

iacore: How do I reconciliate with this?

image

WebReflection: data is a standard CharacterData accessor and it's the fastest one out there for CharacterData instances, namely `#text` and `#comment` ... you can see the implementation in the CharacterData class: https://developer.mozilla.org/en-US/docs/Web/API/CharacterData/data

all other things are **not CharacterData** ... you are screening per `TEXT_NODE` and that **is a CharacterData**.

Use the right tool for the job 😉

WebReflection:
I am not going to Approve and run until this is draft ... please let me know when this is not draft anymore, thanks!

iacore:
So, I discovered another bug that I don't know how to fix. Can you fix it?

WebReflection:
I am not sure I know what bug you are talking about ... but if it has nothing to do with `textContent` it should likely not be part of this PR?

iacore:
fixed it.

PR #279, opened 2024-06-12
Add test for templateElement.content.childNodes[0].replaceWith

iacore:
added a test. the test should not fail.

### DIFF
--- a/test/html/template-element.js
+++ b/test/html/template-element.js
@@ -47,3 +47,11 @@ const docWithTemplateAttribute = parseHTML(`<div template="anything"><p>not insi
 
 assert(docWithTemplateAttribute.querySelector('*').tagName, 'P');
 assert(docWithTemplateAttribute.querySelectorAll('*').length, 1);
+
+
+let {document: document2} = parseHTML('<template><img id="el-14" class="has-svg" src="./favicon.svg" inline=""></template>');
+let template2 = document2.documentElement;
+assert(template2.innerHTML, '<img id="el-14" class="has-svg" src="./favicon.svg" inline="">');
+
+template2.content.childNodes[0].replaceWith(document.createElement('br'))
+assert(template2.innerHTML, '<br>', 'expected content');


test/html/template-element.js
```diff
@@ -47,3 +47,11 @@ const docWithTemplateAttribute = parseHTML(`<div template="anything"><p>not insi
 
 assert(docWithTemplateAttribute.querySelector('*').tagName, 'P');
 assert(docWithTemplateAttribute.querySelectorAll('*').length, 1);
+
+
+let {document: document2} = parseHTML('<template><img id="el-14" class="has-svg" src="./favicon.svg" inline=""></template>');
+let template2 = document2.documentElement;
+assert(template2.innerHTML, '<img id="el-14" class="has-svg" src="./favicon.svg" inline="">');
+
+template2.content.childNodes[0].replaceWith(document.createElement('br'))
```
iacore, line 56: if you save

```
const content = template2.content
```

`content.toString()` will give the correct content.

`template2.content.innerHTML` still won't.

WebReflection: yeah, live nodes are not a thing in here ... and won't be because that's not a goal of this project.

WebReflection: To clarify: I don't know if `content` returns an array of its nodes but if that's the case, that's gonna be the case in here. For everything else, JSDOM got you covered. This project is not a replacement for it, it just enables the read-only things a page would land, everything else really need a good use case for it or it won't happen. Do you want to dig into the `DocumentFragment` hole without affecting performance? Please do, if that's the case, I'll be watching 👍

iacore: what should I do instead?

the idea is to swap out `<a/>` with `<c/>` without affecting anything else.

```
<div>
<a/>
<b/>
</div>
```

iacore: I found the solution. The problem is not about live nodes. 

`template.querySelectorAll` works.

`template.content.querySelectorAll` doesn't. `template.content` **clones each node**.

WebReflection: yeah, `templte.content` does that and it's **not a LIVE NodeList** so you got your answer ... that won't land in here, it's no purpose whatsoever, so if you solved, remove the failing-but-not-really test and let's merge your changes?

iacore:
# On HTMLTemplateElement.content

The following snippet from MDN implies that `templateElement.content` should not clone the node.

```
const templateElement = document.querySelector("#foo");
const documentFragment = templateElement.content.cloneNode(true);
```

I tried it in browser. It works as expected.

[deno-dom](https://github.com/b-fuze/deno-dom) works as expected.


>  Do you want to dig into the DocumentFragment hole without affecting performance? 

I don't think it is possible to do that, given the current implementation seems a bit buggy.

```

class HTMLTemplateElement extends HTMLElement {
  constructor(ownerDocument) {
    super(ownerDocument, tagName);
    const content = this.ownerDocument.createDocumentFragment();
    (this[CONTENT] = content)[PRIVATE] = this;
  }

  get content() {
    if (this.hasChildNodes() && !this[CONTENT].hasChildNodes()) {
      for (const node of this.childNodes)
        this[CONTENT].appendChild(node);
    }
    return this[CONTENT];
  }
}
```

If I understand currently, `get content()` is cached once. Consider this example:

```js
templateElement.content;
templateElement.childNodes.append(document.createElement('div'))
templateElement.content; // this would be stale
```

iacore:
Not sure what do do with this, actually.

browser DOM says `templateElement.childNodes` should be empty, while `templateElement.childNodes`.

linkedom seems to insert to template first, and then on first `.content` transfer to the inner DocumentFragment.

WebReflection:
It’s not complicated, rather about SSR performance where you never want to compute things more than once. May I ask your use case here? If it’s to test please read the non-goals section in the README. If it’s about something else please share, thanks.

P.S. I’m really focusing on other things these days and I won’t be able to help much but if it’s a quick thing I will try to help

iacore:
It's about something else. It's already solved.

https://github.com/dringtech/lume/pull/1/files#diff-7ff608c5d5ab477d4ee84e3dd47fb9700039d91695614bb2e06968e824827094L53

PR #284, opened 2024-08-13
Handle tables and additional block like elements in innerText

esamattis:


esamattis:
Oops, sorry for the noise. Sent this PR too early.

Issue #285, opened 2024-09-17
TreeWalker doesn't work when the root element is a TextNode


kasperskei:
```ts
import {parseHTML} from 'linkedom'
const {document} = parseHTML('<html>text</html>')
document.createTreeWalker(document.documentElement.firstChild)
```

Executing the code results in an error:

```bash
14 | 
15 | import {PRIVATE, END, NEXT} from '../shared/symbols.js';
16 | 
17 | const isOK = ({nodeType}, mask) => {
18 |   switch (nodeType) {
19 |     case ELEMENT_NODE:
              ^
TypeError: Right side of assignment cannot be destructured
      at isOK (/linkedom/esm/interface/tree-walker.js:19:10)
      at new TreeWalker (/linkedom/esm/interface/tree-walker.js:56:11)
      at createTreeWalker (/linkedom/esm/interface/document.js:186:52)
```

WebReflection:
what do you expect from crawling a non element node or why are you accessing `firstChild` instead of `firstElementChild` or why aren't you accessing just `document.documentElement` to walk that page tree?

this feels like a *meh, nobody cared to date* issue to me and my time is extremely constrained these days so ... could you please elaborate more on what's the use case of creating a tree walker out of text or comment nodes? why would you move so much logic to start with in the first place? It's a honest question.

kasperskei:
I’m using **linkedom** for pre-rendering and testing an SPA application. One of the client-side libraries uses `MutationObserver` and `TreeWalker` together. This is valid code, as `TreeWalker` accepts an instance of the `Node` interface as its first argument. I want to reduce the differences between **linkedom** and browsers to get accurate results.

WebReflection:
Up and running, thanks for the PR 👋

PR #286, opened 2024-09-17
Fix TreeWalker when the root element is a TextNode

kasperskei:
Fixes https://github.com/WebReflection/linkedom/issues/285

### DIFF
--- a/cjs/interface/tree-walker.js
+++ b/cjs/interface/tree-walker.js
@@ -43,7 +43,7 @@ class TreeWalker {
       end = documentElement[END];
     }
     const nodes = [];
-    while (next !== end) {
+    while (next && next !== end) {
       if (isOK(next, whatToShow))
         nodes.push(next);
       next = next[NEXT];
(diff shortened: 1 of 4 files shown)


esm/interface/tree-walker.js
```diff
@@ -42,7 +42,7 @@ export class TreeWalker {
       end = documentElement[END];
     }
     const nodes = [];
-    while (next !== end) {
+    while (next !== end && next != null) {
```
kasperskei, line 45: Alternative option
```suggestion
    while (next !== end && next.nodeType !== NODE_END) {
```

cjs/interface/tree-walker.js
```diff
@@ -43,7 +43,7 @@ class TreeWalker {
       end = documentElement[END];
     }
     const nodes = [];
-    while (next !== end) {
+    while (next !== end && next != null) {
```
WebReflection, line 46: `while (next && next !== end)` maybe?

WebReflection:
I think `worker.js` is an artifact so there's no point in providing a diff of such artifact but then again, all looks good and the change feels minimal enough. Let CI provide its results and I'll merge this.

Issue #292, opened 2024-12-12
Unexpected characters `$$` were escaped in scripts


lawvs:
Also related to https://github.com/WebReflection/linkedom/issues/158 https://github.com/WebReflection/linkedom/issues/199

Version: 0.18.5

```js
import { parseHTML } from "linkedom"

const html = parseHTML('<html><script>const test = "$$"</script></html>')
//                                                  ^^

console.log(html.document.toString())
// <html><script>const test = "$"</script></html>
//                             ^
// It should be `$$`, but `$`.
```

Issue #199, opened 2023-03-17
`>` inside scripts get escaped.


hacknlove:
```
> const { DOMParser } = require("linkedom");
undefined
> const parser = new DOMParser();
undefined
> parser.parseFromString(`<html><body><script>(() => {if (5>4) { console.log('hello') }})()</script></body></html>`).toString()
`<?xml version="1.0" encoding="utf-8"?><html><body><script>(() =&gt; {if (5&gt;4) { console.log('hello') }})()</script></body></html>`
```

It should be `() => {` and `5>4`

I'm using linkedom@0.14.25

WebReflection:
there are nodes where no escape should happen: title, textarea, and script or css, among others ... this keeps coming up and I've no idea why ... will have a look whenever I'll find time

sdvg:
I have a similar problem, but not sure if it's exactly the same. It's also about odd replacements in script tags:

```
const { document } = parseHTML(`<script>new RegExp('(^something.*$')</script>`)

console.log(document.toString());
```

prints:
```
<script>new RegExp('(^something.*/script>)</script>
```

If it's not the same issue, I'm happy to create a new one ✌️

Issue #158, opened 2022-09-09
Text escaping of $& becomes ><


SvanteRichter:
Hello!

I think I have found a slight issue with how text is escaped. using this code:

```js
import { parseHTML } from "https://esm.sh/linkedom"
const { document } = parseHTML('')
document.appendChild(document.createElement('script'))
document.querySelector("script").textContent = 'var $=1;$&&true'
console.log(document.toString())
```

logs `<script>var $=1;><&true</script>` as output. `$&` has been replaced by `><`.

If I run similar code in a browser I get `<script>var $=1;$&&true</script>`. The code I ran in the browser is

```js
var doc = document.createDocumentFragment()
doc.appendChild(document.createElement('script'))
doc.querySelector("script").textContent = 'var $=1;$&&true';
console.log(doc.querySelector("script").outerHTML)
```

This is a bit problematic for me since it means that inlineing my JS after SSR produces invalid code which will not run.

I have tested it in deno (v1.24.3), node (v18.2.0) and bun.sh (v0.1.11) so I don't think it's anything platform specific.

Thanks!

WebReflection:
what's `document.appendChild` supposed to do there? you have a body, a head, why document to append scripts? just to be sure it's not an issue caused yet another time by empty documents to parse

WebReflection:
those are two different code paths/logic ... one thing is `textContent = '...'` which is not changed in code (for what I could tell) the other is parseHTML with stuff that the parser *might* get wrong so it's a 3rd party issue ... which one is true? both?

the toString actuallt escape chars, don't drop these https://github.com/WebReflection/linkedom/blob/main/esm/interface/text.js#L41

PR #293, opened 2024-12-13
fix: unexpected special replacement pattern

lawvs:
Fix https://github.com/WebReflection/linkedom/issues/292 https://github.com/WebReflection/linkedom/issues/158 https://github.com/WebReflection/linkedom/issues/199

See also https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement

photo

### DIFF
--- a/cjs/html/text-element.js
+++ b/cjs/html/text-element.js
@@ -10,7 +10,11 @@ class TextElement extends HTMLElement {
 
   toString() {
     const outerHTML = toString.call(this.cloneNode());
-    return outerHTML.replace(/></, `>${this.textContent}<`);
+    // Use `join` in here to prevent unexpected special replacement patterns
+    // See also 
+    // https://github.com/WebReflection/linkedom/issues/292
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement
+    return outerHTML.split('><').join(`>${this.textContent}<`);
   }
 }
 exports.TextElement = TextElement
--- a/cjs/html/text-element.js
+++ b/cjs/html/text-element.js
@@ -10,11 +10,7 @@ class TextElement extends HTMLElement {
 
   toString() {
     const outerHTML = toString.call(this.cloneNode());
-    // Use `join` in here to prevent unexpected special replacement patterns
-    // See also 
-    // https://github.com/WebReflection/linkedom/issues/292
-    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement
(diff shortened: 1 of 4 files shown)


WebReflection:
How bizarre… I completely forgot about it, nice catch but split join feels unnecessarily expensive? Although I’m not sure doubling $ in text content would be faster or better, have you tried that? 🤔 all $\S should become $$\S

WebReflection:
Another way could be using a one-off arrow that should not have special meanings 🤔

cjs/html/text-element.js
```diff
@@ -10,7 +10,11 @@ class TextElement extends HTMLElement {
 
   toString() {
     const outerHTML = toString.call(this.cloneNode());
-    return outerHTML.replace(/></, `>${this.textContent}<`);
+    // Use `join` in here to prevent unexpected special replacement patterns
+    // See also 
+    // https://github.com/WebReflection/linkedom/issues/292
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement
+    return outerHTML.split('><').join(`>${this.textContent}<`);
```
WebReflection, line 17: I would just add () => in front of the replacement, it feels faster and easier on both RAM and GC

cjs/html/text-element.js
```diff
@@ -10,7 +10,11 @@ class TextElement extends HTMLElement {
 
   toString() {
     const outerHTML = toString.call(this.cloneNode());
-    return outerHTML.replace(/></, `>${this.textContent}<`);
```
WebReflection, line 13:
```js
return outerHTML.replace(/></, () => `>${this.textContent}<`);
```

esm/html/text-element.js
```diff
@@ -9,6 +9,6 @@ export class TextElement extends HTMLElement {
 
   toString() {
     const outerHTML = toString.call(this.cloneNode());
-    return outerHTML.replace(/></, `>${this.textContent}<`);
```
WebReflection, line 12:
```js
return outerHTML.replace(/></, () => `>${this.textContent}<`);
```

worker.js
```diff
@@ -8567,7 +8567,7 @@ class TextElement extends HTMLElement {
 
   toString() {
     const outerHTML = toString.call(this.cloneNode());
-    return outerHTML.replace(/></, `>${this.textContent}<`);
```
WebReflection, line 8570:
```js
return outerHTML.replace(/></, () => `>${this.textContent}<`);
```

lawvs:
> Another way could be using a one-off arrow that should not have special meanings 🤔

I created a simple bench, and the results show that replace indeed has better performance without using regex.

image

WebReflection:
It's up and running (or better, published)

PR #302, opened 2025-04-26
Trim the result of URLs in "href".

### LINKED ISSUE #301: Trimming of "href" results
I noticed that there are websites that include newlines and leading or trailing whitespaces in canonical links. Linkedom does not automatically trim the results. Here is an example:

```
import { parseHTML } from 'linkedom';

const html = `<!DOCTYPE html><html><head><title>Test</title><link rel='canonical' href="		https://example.com/
	"></head><body></body></html>`;
const { document } = parseHTML(html);
console.log(`canonical link: (not trimmed) length=${document.querySelector('html > head > link[rel="canonical"]').href.length}`);
console.log(`canonical link: trimmed length=${document.querySelector('html > head > link[rel="canonical"]').href.trim().length}
```

My understanding is that according to the spec, it should be trimmed:

https://url.spec.whatwg.org/#dom-url-href is quite long, but states:
> The protocol setter steps are to basic URL parse the given value, followed by U+003A (:), with this’s URL as url and scheme start state as state override.

And the mentioned ["basic URL parse"](https://url.spec.whatwg.org/#concept-basic-url-parser):
> Remove any leading and trailing C0 control or space from input.

Firefox and Chrome also trim.

I can take a look and create a pull request. But I wanted to confirm first that trimming is the desired results in Linkedom for this example.

philipp-classen:
fixes #301

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -15,9 +15,9 @@ jobs:
         node-version: [20]
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
     - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v2
+      uses: actions/setup-node@v4
       with:
         node-version: ${{ matrix.node-version }}
         cache: 'npm'
--- a/cjs/html/anchor-element.js
+++ b/cjs/html/anchor-element.js
@@ -15,7 +15,7 @@ class HTMLAnchorElement extends HTMLElement {
   }
 
   /* c8 ignore start */ // copy paste from img.src, already covered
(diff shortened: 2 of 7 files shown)


WebReflection:
awesome, thanks ... do you mind changing also the checkout vo `v4` in here https://github.com/WebReflection/linkedom/blob/main/.github/workflows/node.js.yml#L18-L20 so that we can see CI happy? You could also bump node to 22 but I want to see green lights before merging and apparently `v2` checkout just recently broke everything.

philipp-classen:
Updated to v4. The build is green now.

WebReflection:
@philipp-classen it's up and running 🥳

PR #304, opened 2025-05-29
Make attribute value and text content escaping more conforming

### LINKED ISSUE #216: Characters (&), (<) and (>) escaped incorrectly in xml attributes
```
The ampersand character (&) and the left angle bracket (<) MUST NOT appear in their literal form, [...] they MUST be [escaped](https://www.w3.org/TR/xml11/#dt-escape) using [...] the strings "&amp;" and "&lt;" respectively. The right angle bracket (>) may be represented using the string "&gt;", and MUST, [for compatibility](https://www.w3.org/TR/xml11/#dt-compat), be escaped using either "&gt;"
```
Full source : https://www.w3.org/TR/xml11/#syntax

Test snippet: 

```
const document = (new DOMParser).parseFromString('<root attr="&lt;&gt;&amp;"></root>', 'text/xml');

assert(document.toString(), '<?xml version="1.0" encoding="utf-8"?><root attr="&lt;&gt;&amp;" />');
```

Expected: `<?xml version="1.0" encoding="utf-8"?><root attr="&lt;&gt;&amp;" />`
Got instead: `<?xml version="1.0" encoding="utf-8"?><root attr="<>&" />`

PolariTOON:
Fixes #216
Supersedes and closes #217

The HTML spec defines which characters to escape when [serializing HTML](https://html.spec.whatwg.org/multipage/parsing.html#escapingString). This got changed recently to also serialize `<` and `>` in attribute values.

This aligns with the DOM Parsing and Serialization spec which defines its own rule for [escaping attribute values](https://w3c.github.io/DOM-Parsing/#dfn-serializing-an-attribute-value) and for [escaping text nodes](https://w3c.github.io/DOM-Parsing/#dfn-xml-serializing-a-text-node), both requiring these two characters to always be escaped.

HTML also requires non-breaking spaces to be escaped (in HTML).
Also browsers de facto serialize tabs, line feeds and carriage returns in attribute values in XML (there is no spec for this, but there is a [web platform test](https://github.com/web-platform-tests/wpt/blob/master/domparsing/XMLSerializer-serializeToString.html#L155)). JSDOM does this too.

This PR implements those requirements.

This PR also brings another fix about `<title>` elements which used to be considered a special case when serializing in this codebase, while i couldn't find anything in the spec about this (the `<title>` element is not in [this list](https://html.spec.whatwg.org/multipage/parsing.html#serialising-html-fragments:text)). Instead, we customize its `innerHTML` setter to escape everything, which prevents child elements to be added to them through this property (for HTML documents only).

### DIFF
--- a/cjs/html/title-element.js
+++ b/cjs/html/title-element.js
@@ -1,17 +1,22 @@
 'use strict';
 const {registerHTMLClass} = require('../shared/register-html-class.js');
+const {escapeHtmlTextContent} = require('../shared/text-escaper.js');
+const {ignoreCase} = require('../shared/utils.js');
 
-const {TextElement} = require('./text-element.js');
+const {HTMLElement} = require('./element.js');
 
 const tagName = 'title';
 
 /**
  * @implements globalThis.HTMLTitleElement
  */
-class HTMLTitleElement extends TextElement {
+class HTMLTitleElement extends HTMLElement {
   constructor(ownerDocument, localName = tagName) {
     super(ownerDocument, localName);
   }
+
+  get innerHTML() { return super.innerHTML; }
+  set innerHTML(html) { super.innerHTML = ignoreCase(this) ? escapeHtmlTextContent(html) : html; }
 }
 
 registerHTMLClass(tagName, HTMLTitleElement);
--- a/cjs/interface/attr.js
+++ b/cjs/interface/attr.js
@@ -4,14 +4,12 @@ const {CHANGED, VALUE} = require('../shared/symbols.js');
 const {String, ignoreCase} = require('../shared/utils.js');
 const {attrAsJSON} = require('../shared/jsdon.js');
 const {emptyAttributes} = require('../shared/attributes.js');
+const {escapeHtmlAttributeValue, escapeXmlAttributeValue} = require('../shared/text-escaper.js');
 
 const {attributeChangedCallback: moAttributes} = require('./mutation-observer.js');
 const {attributeChangedCallback: ceAttributes} = require('./custom-element-registry.js');
 
 const {Node} = require('./node.js');
-const {escape} = require('../shared/text-escaper.js');
-
-const QUOTE = /"/g;
 
 /**
  * @implements globalThis.Attr
@@ -46,7 +44,7 @@ class Attr extends Node {
     if (emptyAttributes.has(name) && !value) {
       return ignoreCase(this) ? name : `${name}=""`;
     }
-    const escapedValue = (ignoreCase(this) ? value : escape(value)).replace(QUOTE, '&quot;');
+    const escapedValue = ignoreCase(this) ? escapeHtmlAttributeValue(value) : escapeXmlAttributeValue(value);
     return `${name}="${escapedValue}"`;
   }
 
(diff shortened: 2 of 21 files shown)


WebReflection:
to clarify, this is a perfectly valid `document.title` every browser will understand:

```html
<title>1 <and>&</and> 2</title>
```

it's a special node that acts like `style` or `script` (a data node) and I am not sure it needs fixing as it is right now ... true that if *html-escape* should consider non breaking spaces too I need to fix [this library](https://www.npmjs.com/package/html-escaper) too and these changes will be welcome but it's weird nobody ever complained to date about that library not catching the `&nbsp;` explicitly which is why I think this might be a breaking thing in here ... do you have concrete examples where the current state breaks the Web? Thank you!

PolariTOON:
My general use case for the three PRs i have opened is actually the same one. I'm doing SSR of XHTML and SVG documents and until now i was indeed using JSDOM for this. However i'm starting to face performance issues, as JSDOM implements many things i don't need. Basically all i need is parsing, basic DOM manipulation (setting attributes and adding nodes) and serializing the whole thing, and LinkeDOM seems to fit my needs. Thanks for making it!

The data i generate is versioned though, so while this not about testing, i indeed want to minimize the diffs, and together, the PRs i have opened achieve avoiding having any diff on my data.

That said, the particular issues i'm trying to fix with this PR are:
- avoiding generating malformed XHTML documents that browsers can't open, due to `<` and `>` not being escaped in attribute values.
- making attribute values containing tabs and newlines eventually roundtrip without data loss. XML parsers in browsers normalize unescaped tabs and line feeds as spaces. I'm displaying attributes with CSS, so i want to keep these characters and that requires escaping them, as specced.
- avoiding some XSS issues with the current `<title>` implementation, that i realized while looking at [this test](https://github.com/WebReflection/linkedom/blob/main/test/html/document.js#L48):
  ```js
  const {document} = parseHTML("<html><head><title>Title</title></head></html>");
  document.title = "</title><script>alert(1)</script>";
  // or document.querySelector("title").textContent = "</title><script>alert(1)</script>";
  // or document.querySelector("title").innerHTML = "</title><script>alert(1)</script>";
  document.toString(); // <html><head><title></title><script>alert(1)</script></title></head></html>
  ```
  This PR fixes the three XSS issues above by:
  - not overriding `titleElement.toString()`
  - not delegating `titleElement.innerHTML` to `titleElement.textContent`
  - escaping values provided to `titleElement.innerHTML`'s setter
  
  That last item is precisely the special-case you're talking about in the HTML parser: most of the time you don't have to (double) escape the title content yourself as the engine (LinkeDOM here) is supposed to do it for you.
  Should i add this example as testcases?

While looking at the spec, i also fixed other escaping cases i encountered. `&nbsp;` is just one of these so I don't have any use case for it to be honest (as i'm focusing on XHTML rather than on HTML).

WebReflection:
> JSDOM implements many things i don't need

this is the key and reason this project is fast ... so let's talk about this: why would you need attributes insertion order, as example or why should I add complexity for things nobody practically cares about?

I am afraid I'll be a push back for anything not really interesting, relevant, solving anything useful and so on ... there is a bug? PR welcome ... there is a not 100% strict behavior nobody cares about or use in the real world? PR **not welcome** as I don't care maintaining those things and neither should you.

PR #306, opened 2025-05-29
Preserve attribute insertion order

PolariTOON:
d8c3420ba1e3a285fdef11bd637a25eb173ebc77 made it so attributes are stored in insertion order when parsing. This PR makes it so insertion order is also preserved when adding a new attribute afterwards by manipulating the DOM with setters or methods such as `setAttribute()`.

Unfortunately this increases the `setAttribute` complexity. Could you provide some guidance to avoid introducing the loop i have currently added? Does a pointer to the first child element of a parent element exist for example? Or is it fine like this? (as you said in the readme, "everything is a while(...) loop away")


WebReflection:
Similar question: what issue is this solving?

PolariTOON:
This is the de facto behavior of browsers so i was kind of expecting that (even if the spec [does not](https://html.spec.whatwg.org/multipage/parsing.html#serialising-html-fragments:implementation-defined) require that) and that would ease migrating from JSDOM to LinkeDOM in my case (versioned SSR-generated code). So this is just about consistency within this codebase and with other implementations. I don't have a better reason to give, sorry :sweat_smile:

WebReflection:
I won’t work on this so PR welcomed with benchmark attached

WebReflection:
there is already a benchmark folder https://github.com/WebReflection/linkedom/tree/main/test/benchmark so that PR should compare, within the same HW and conditions, the before/after changes so we can revalidate in other devices and eventually move forward.

PR #307, opened 2025-05-29
Fix insertAdjacentHTML and createContextualFragment not working on non-HTML documents

Gudine:
The functions `Element.insertAdjacentHTML` and `Range.createContextualFragment` resulted in crashes when used in XML documents, as they both rely on functionality from the `template` HTML element, which only works in HTML documents. I fixed the issue by adapting the logic used in `HTMLTemplateElement` into a separate utility function.

To test this, I've moved some tests from `html/element.js` to `interface/element.js`, as they don't apply just to HTML elements.

I'm not sure how feasible it would be, and it is out of scope of this PR, but it might also make sense to make the interface tests run on all three document types, as such tests would've caught both errors fixed here.

### DIFF
--- a/esm/shared/utils.js
+++ b/esm/shared/utils.js
@@ -38,3 +38,14 @@ export const setAdjacent = (prev, next) => {
   if (next)
     next[PREV] = prev;
 };
+
+export const htmlToFragment = (ownerDocument, html) => {
+  const fragment = ownerDocument.createDocumentFragment();
+
+  const elem = ownerDocument.createElement('');
+  elem.innerHTML = html;
+
+  for (const node of elem.childNodes) fragment.appendChild(node.cloneNode(true));
+
+  return fragment;
+};
--- a/cjs/interface/element.js
+++ b/cjs/interface/element.js
@@ -21,6 +21,7 @@ const {
 } = require('../shared/symbols.js');
 
 const {
+  htmlToFragment,
   ignoreCase,
   knownAdjacent,
   localCase,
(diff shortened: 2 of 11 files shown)


esm/shared/utils.js
```diff
@@ -38,3 +38,14 @@ export const setAdjacent = (prev, next) => {
...
     next[PREV] = prev;
 };
+
+export const htmlToFragment = (ownerDocument, html) => {
+  const fragment = ownerDocument.createDocumentFragment();
+
+  const elem = ownerDocument.createElement('');
+  elem.innerHTML = html;
+
+  for (const node of elem.childNodes) fragment.appendChild(node.cloneNode(true));
```
WebReflection, line 48: why `cloneNode(true)` is needed in here? It's also worth mentioning this is a classic *move nodes* operations which is *O(1)* operation changing the `PREV` and `NEXT` links to the first and last child + updating `PARENT` ... mind refactoring this to make it faster? If needed some hint around *how to* check what `appendChild` does internally

Gudine:
Thanks for the suggestion! I implemented it to the best of my understanding, though it is still O(n) because of the parent loop.

Also, in the process I added JSDoc to the function, which ended up changing the type of `createContextualFragment`, but I can remove it if necessary.

WebReflection:
>  it is still O(n) because of the parent loop.

there is *O(n)* and *O(n)* but the whole point in there is that you don't need to bloat RAM and pressure GC with clones and logic that is *one shot* away so this latest change is the way to go in *LinkeDOM*, if I took the *everything is O(n) anyway* approach in here it would be deadly slow, I hope this follow up makes sense to you ... I'll try to find time to review, check/test and publish this MR today or tomorrow.

WebReflection:
@Gudine it's up and running: thank you 🥳

Issue #327, opened 2026-06-17
`el.style.property = ''` and `e.style.setProperty('property', '')` should delete the property


pucelle:
As title said,
`el.style.property = ''` and `e.style.setProperty('property', '')` should delete the property.

WebReflection:
YAGNI in here?

PR #328, opened 2026-07-05
fix: reading an unset style property returns "" per CSSOM, not undefined

magicismight:
## Problem

Reading a style property that isn't set returns `undefined`, but per [CSSOM](https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-getpropertyvalue) both the camel-cased property getter and `getPropertyValue()` must return the **empty string** for an unset property.

```js
import { parseHTML } from "linkedom";
const { document } = parseHTML("<!doctype html><html><body></body></html>");
const el = document.createElement("p");

el.style.color;                         // => undefined   (browser: "")
el.style.getPropertyValue("color");     // => undefined   (browser: "")
```

Because the property getter is routed through the same handler as `getPropertyValue`, both are affected. Code that assumes the spec — e.g. `el.style.color.replaceAll(...)` or any string method — throws `Cannot read properties of undefined` on an element that simply hasn't set that property. This bites libraries that parse arbitrary HTML and read inline styles (ProseMirror's DOMParser, etc.).

## Fix

Normalize an absent property to `""` in the style getter (`esm/interface/css-style-declaration.js`), which also covers `getPropertyValue`. Setting/removing/serialization are untouched; a set property still returns its value, and `cssText` is unchanged.

```js
const value = style.get(uhyphen(name));
return value === void 0 ? '' : value;
```

CJS is regenerated via `npm run cjs`. A test is added in `test/interface/css-style-declaration.js`; the full suite (`node test/index.js`) passes.

## Not a duplicate

Distinct from the open CSS issues:
- #318 replaces the CSSOM library for parsing `<style>` sheets (`element.sheet`), not the inline `element.style` declaration.
- #327 concerns *setting* an empty string (delete semantics), not *reading* an unset property.

### DIFF
--- a/cjs/interface/css-style-declaration.js
+++ b/cjs/interface/css-style-declaration.js
@@ -36,7 +36,10 @@ const handler = {
       return getKeys(style).length;
     if (/^\d+$/.test(name))
       return getKeys(style)[name];
-    return style.get(uhyphen(name));
+    // Per CSSOM, reading a property that is not set returns the empty string,
+    // not undefined (which also matches getPropertyValue, routed here).
+    const value = style.get(uhyphen(name));
+    return value === void 0 ? '' : value;
   },
 
   set(style, name, value) {
--- a/cjs/interface/css-style-declaration.js
+++ b/cjs/interface/css-style-declaration.js
@@ -36,10 +36,7 @@ const handler = {
       return getKeys(style).length;
     if (/^\d+$/.test(name))
       return getKeys(style)[name];
-    // Per CSSOM, reading a property that is not set returns the empty string,
-    // not undefined (which also matches getPropertyValue, routed here).
-    const value = style.get(uhyphen(name));
-    return value === void 0 ? '' : value;
+    return style.get(uhyphen(name)) ?? '';
   },
 
   set(style, name, value) {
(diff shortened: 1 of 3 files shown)


cjs/interface/css-style-declaration.js
```diff
@@ -36,7 +36,10 @@ const handler = {
       return getKeys(style).length;
     if (/^\d+$/.test(name))
       return getKeys(style)[name];
-    return style.get(uhyphen(name));
```
WebReflection, line 39: I'd rather go for:

```js
return style.get(uhyphen(name)) ?? '';
```

esm/interface/css-style-declaration.js
```diff
@@ -35,7 +35,10 @@ const handler = {
       return getKeys(style).length;
     if (/^\d+$/.test(name))
       return getKeys(style)[name];
-    return style.get(uhyphen(name));
+    // Per CSSOM, reading a property that is not set returns the empty string,
+    // not undefined (which also matches getPropertyValue, routed here).
+    const value = style.get(uhyphen(name));
+    return value === void 0 ? '' : value;
```
WebReflection: I'd rather go for:

```js
return style.get(uhyphen(name)) ?? '';
```

WebReflection:
see comments, otherwise I am OK with this, thanks.

cjs/interface/css-style-declaration.js
magicismight: Done, thanks.

WebReflection:
up and running, thank you!

PR #331, opened 2026-07-14
fix: validate parent node mutations

ShiroKSH:
## Summary

- Reject parent-node mutations whose reference or replaced node belongs to another parent.
- Reject inserting an ancestor into a descendant, preventing cyclic trees.
- Preserve the existing detached insertBefore(node, node) no-op.

## Tests

- npm test
- Worker bundle repro for an invalid foreign reference

### DIFF
--- a/cjs/mixin/parent-node.js
+++ b/cjs/mixin/parent-node.js
@@ -29,6 +29,13 @@ const {nextElementSibling} = require('./non-document-type-child-node.js');
 
 const isNode = node => node instanceof Node;
 
+const validate = (parentNode, node, before) => {
+  if (before && before !== parentNode[END] && before.parentNode !== parentNode)
+    throw new Error('node is not a child');
+  if (node === parentNode || node.contains(parentNode))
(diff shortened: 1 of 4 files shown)


WebReflection:
a bit defensive which is a bit of a slippery slope or not the goal of this project or goodbye performance but it seems fair. may I ask what's the use case for this, beside stricter adherence with DOM standard which is **not** a goal of this project?

ShiroKSH:
Fair point. I do not have a production use case.

I found it while testing invalid mutation paths. The failure is not just a spec mismatch. A foreign reference can put a node in one parent list while its parentNode points to another. replaceChild can remove a node from the wrong parent. An ancestor insertion can create a cycle and make traversal hang.

If invalid input is out of scope then this is not worth the hot path cost. Happy to close the PR.

WebReflection:
my time is limited and logic in the hot path for something intentionally broken is probably not worth a fix, imho. Closing but thanks for expanding on that.

PR #333, opened 2026-09-20
Fix innerHTML/innerText to coerce non-string values

### LINKED ISSUE #308: Exceptions thrown when setting Element's innerHTML or innerText with non-string value
Reproduciable example:

```js
import { parseHTML } from 'linkedom';

const window = parseHTML('<!DOCTYPE html><html><head></head><body></body></html>');

// These are OK:
// window.document.body.innerHTML = '1';
// window.document.body.innerText = '1';

// These are not:
window.document.body.innerHTML = 1;
window.document.body.innerText = 1;
```

The behavior tested on the browser seems to be consistent with

```js
set innerHTML(value) {
  internalSetInnerHTMLWithString(this, value === null ? '' : String(value));
}
```

dyk1454683243-sudo:
Fixes #308

Setting `innerHTML` or `innerText` with a non-string (for example `1` or `null`) currently throws. Browsers treat these accessors as `[LegacyNullToEmptyString] DOMString`: `null` becomes `''`, and any other value is stringified.

This keeps the change on the setter path only:
- `setInnerHtml` coerces with `html === null ? '' : String(html)` before `parseFromString`
- `Element` now has a matching `set innerText` that uses the same coerce and replaces children like `textContent` (the existing `get innerText` whitespace rules are unchanged)
- Regression tests cover `innerHTML = 1`, `innerHTML = null`, and `innerText = 1`

```js
document.body.innerHTML = 1;    // '1', no throw
document.body.innerHTML = null; // ''
document.body.innerText = 1;    // visible text '1'
```


dyk1454683243-sudo:
Withdrawing this PR while I clean up a high-volume open-PR backlog. Sorry for the noise — happy to come back later with a focused change if useful.

PR #334, opened 2026-09-20
Treat DOMTokenList.toggle undefined force as omitted

### LINKED ISSUE #277: In 'HTMLElement.classList.toggle' the 'force' attribute works uncorrectly.
```typescript
export const toggle = <T extends HTMLElement> (className:string, force?:boolean) => (el:T) : boolean => el.classList.toggle(className, force)

toggle(className)(el)
```

In this example, if the 'force' attribute is not passed to the 'toggle' function, 'linkedom' will interpret it as being passed an attribute equal to 'undefined' and identify it as a 'false' value.

The correct way is for 'linkedom' to ignore the 'force' attribute if it is 'undefined'.

Everything works fine in the browser.

Аnd also in 'linkedom' everything works correctly if I do this check myself:

```typescript
export const toggle = <T extends HTMLElement> (className:string, force?:boolean) => (el:T) : boolean => force === undefined ? el.classList.toggle(className) : el.classList.toggle(className, force)
```

dyk1454683243-sudo:
Fixes #277

`DOMTokenList.toggle(token, force)` now treats `force === undefined` like an omitted second argument (plain toggle), matching browsers and Web IDL optional boolean.

- Missing token + `undefined` force → add (same as one-arg `toggle`)
- Present token + `undefined` force → remove
- `force` `true` / `false` still force-add / force-remove

Focused change in `esm/dom/token-list.js` (CJS mirrored) plus regression tests.


dyk1454683243-sudo:
Withdrawing this PR while I clean up a high-volume open-PR backlog. Sorry for the noise — happy to come back later with a focused change if useful.

